PR #5, opened 2014-09-22
vendor dependencies using Godep

btc:
addresses: https://github.com/jbenet/datastore.go/issues/3


jbenet:
sigh. vendoring has such a stupid cost right now. can't wait to have IPFS.

PR #8, opened 2014-11-03
Updated leveldb

jbenet:


jbenet:
@maybebtc ahhhhh it was elastigo that was broken and needed to be an earlier version.

jbenet:
I'll merge this in to be up to date.

PR #9, opened 2014-11-16
feat(interface) ThreadSafeDatastoreCloser

btc:
Expose `Close()` to make it possible to clean up LevelDB instances.

License: MIT
Signed-off-by: Brian Tiger Chow brian@perfmode.com

### DIFF
--- a/datastore.go
+++ b/datastore.go
@@ -2,6 +2,7 @@ package datastore
 
 import (
 	"errors"
+	"io"
 )
 
 /*
@@ -61,6 +62,13 @@ type Datastore interface {
 type ThreadSafeDatastore interface {
 	Datastore
 	IsThreadSafe()
+}
+
+// ThreadSafeDatastoreCloser extends the ThreadSafeDatastore with the io.Closer
+// interface
+type ThreadSafeDatastoreCloser interface {
+	ThreadSafeDatastore
+	io.Closer
 }
 
 // Errors
--- a/leveldb/datastore.go
+++ b/leveldb/datastore.go
@@ -13,7 +13,7 @@ type Datastore struct {
 
 type Options opt.Options
 
-func NewDatastore(path string, opts *Options) (ds.ThreadSafeDatastore, error) {
+func NewDatastore(path string, opts *Options) (ds.ThreadSafeDatastoreCloser, error) {
 	var nopts opt.Options
 	if opts != nil {
 		nopts = opt.Options(*opts)
--- a/sync/sync.go
+++ b/sync/sync.go
@@ -16,7 +16,7 @@ type MutexDatastore struct {
 
 // MutexWrap constructs a datastore with a coarse lock around
 // the entire datastore, for every single operation
-func MutexWrap(d ds.Datastore) ds.ThreadSafeDatastore {
+func MutexWrap(d ds.Datastore) ds.ThreadSafeDatastoreCloser {
 	return &MutexDatastore{child: d}
 }
 
@@ -62,3 +62,5 @@ func (d *MutexDatastore) KeyList() ([]ds.Key, error) {
 	defer d.RUnlock()
 	return d.child.KeyList()
 }
+
+func (d *MutexDatastore) Close() error { return nil } // no-op


datastore.go
```diff
@@ -63,6 +64,13 @@ type ThreadSafeDatastore interface {
 	IsThreadSafe()
 }
 
+// ThreadSafeDatastoreCloser extends the ThreadSafeDatastore with the io.Closer
+// interface
+type ThreadSafeDatastoreCloser interface {
+	ThreadSafeDatastore
+	io.Closer
```
jbenet, line 71: I want to keep the `Closer` interface outside of the `Datastore*` interfaces. I think the leveldb instance (or whatever you want the `Close` func in) can just conform to `io.Closer`.

btc, line 71: LevelDB returns a ThreadSafeDatastore. Prefer to expose the concrete struct?

jbenet:
@maybebtc i think the `Closer` interface doesnt belong in the datastore interfaces themselves. I think the concrete types (e.g. `leveldb.Datastore`) can just implement Closer where needed.

btc:
I'll have the LevelDB constructor return a concrete instance then.

jbenet:
@maybebtc yeah that makes sense. feel free to break up `leveldb.Datastore` into two parts: an interface, and an unexported concrete type

jbenet:
``` Go
// Datastore wraps a levelDB instance with a Datastore interface
type Datastore interface {
    ds.ThreadSafeDatastore
    io.Closer

    DB() *leveldb.DB
}

// datastore uses a standard Go map for internal storage.
type datastore struct {
    db *leveldb.DB
}
```

btc:
closed in favor of https://github.com/jbenet/go-datastore/pull/10

btc:
@jbenet 

One minor note...

Since the datastore interfaces to not expose Close(), the client is left to implement her own wrapper. This is perfectly fine, but it might be nice for the library to anticipate this need and provide this behavior just as it does with the mutex wrapper.

Example:

``` go
// ClientAppStore returns a Datastore that may need to be closed in Production
func ClientAppStore(in_test_env bool) MyThreadSafeDatastoreCloser {
  if in_test_env {
    return ds.NewMapDatastore() // compiler error
  }
  return leveldb.NewDatastore()
}
```

To fix this, each go-datastore client must implement a trivial Wrapper around datastores.

``` go
type ThreadSafeDatastoreCloserWrapper struct {
  ds.ThreadSafeDatastore
}

func (w *ThreadSafeDatastoreCloserWrapper) Close() error {
return nil // no-op
}
```

For now, I'll implement a wrapper for my specific use case.

jbenet:
@maybebtc i hear you. It's a tough call. A few things to bear in mind:
1. I'm worried by interface composition explosion. Every datastore sub-interface composing with all others would quickly yield rote work in writing out permutations, and awful `ThreadSafeCloserNonBlockingSymlinkDatastore` javarrhea. I dont see this as a viable route.
2. `Close` is a common enough thing that it _may_ be ok to expand the datastore interface with. So if we _reaaaally_ want it, i'd rather implement explicit nil `Close` funcs than permuations.
3. the `ThreadSafeDatastore` is a different case, in that there is a wrapper that implements more functionality around non-`ThreadSafeDatastore`s, and there is an interface datastores can signal when they implement it internally. (perhaps i shouldn't have put it inside the main `Datastore` lib, but inside the `datastore/sync` pkg...)
4. This is not as nice in languages with inheritance, but in Go, the user maybe should be specifying the interface of her own datastores explicitly. I think it's easier to pick and choose what you want all your datastores to conform to:
   
   ``` Go
   // in go-ipfs/core/datastore.go
   
   // Datastore is the common interface all our datastore backends must conform to
   type Datastore interface {
     ds.Datastore
     syncds.ThreadSafeDatastore
     ktds.KeyTransformDatastore
     symds.SymlinkDatastore
     io.Closer
     sync.Locker
     log.Loggable
   }
   ```

So i think for now let's keep as is? We can revisit this if a better strategy is found.

btc:
I agree with all points. There's no harm in being careful about over-extending the library.

as an aside, here's a link to the wrapper: https://github.com/jbenet/go-ipfs/blob/feat/daemon-init/util/datastore_closer.go

PR #14, opened 2015-03-11
Fs cleanup

tv42:
The fs backend module is problematic --  see https://github.com/datastore/datastore/issues/28 -- but I couldn't resist fixing this glaring bug.

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -13,7 +13,7 @@ import (
 
 var ObjectKeySuffix = ".dsobject"
 
-// Datastore uses a standard Go map for internal storage.
+// Datastore uses a uses a file per key to store values.
 type Datastore struct {
 	path string
 }
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -29,7 +29,7 @@ func NewDatastore(path string) (ds.Datastore, error) {
 
 // KeyFilename returns the filename associated with `key`
 func (d *Datastore) KeyFilename(key ds.Key) string {
-	return filepath.Join(d.path, key.String(), ObjectKeySuffix)
+	return filepath.Join(d.path, key.String()+ObjectKeySuffix)
 }
 
 // Put stores the given value.
--- a/fs/fs_test.go
+++ b/fs/fs_test.go
@@ -2,6 +2,8 @@ package fs_test
 
 import (
 	"bytes"
+	"io/ioutil"
+	"path"
 	"testing"
 
 	. "github.com/jbenet/go-datastore/Godeps/_workspace/src/launchpad.net/gocheck"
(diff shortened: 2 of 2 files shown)


fs/fs.go
```diff
@@ -29,7 +29,7 @@ func NewDatastore(path string) (ds.Datastore, error) {
 
 // KeyFilename returns the filename associated with `key`
 func (d *Datastore) KeyFilename(key ds.Key) string {
-	return filepath.Join(d.path, key.String(), ObjectKeySuffix)
```
jbenet, line 32: oh this intended. I had it your way before, but the current way works like an `index.html` file. So:

```
ds.Put("/foo", "foo")
ds.Put("/foo/bar", "bar")
```

would make

```
datastore
└── foo
    └── .object
    └── bar
        └── .object
```

so we can `mv datastore/foo datastore/baz` and change all the keys `/foo* -> /baz*`.

jbenet, line 32: is there a case where this breaks down horribly?

tv42, line 32: `ds.Put("/foo/.object", "blah")` -- what now? Can't mkdir `foo/.object` as it's already a file.

You could give every directory level an extension too, but that gets ugly, and stops you from doing that `os.MkdirAll`.

The reason I thought this to be unintended is mostly that it wastes at least two inodes per object, and causes way more syscalls. It may be fine for what it is, but high performance it ain't.

jbenet, line 32: yeah, hiding objects is a hack. :/  -- giving dirs ext may be ok. 

Either way-- if think it'll impact perf significantly let's keep it outside. 

We still have to give dirs extensions (or some other workaround) because: 

```
ds.Put("/foo", "foo")
ds.Put("/foo.object/bar", "ha!")
```

:(

tv42, line 32: The real fix is encoding every segment, e.g. hex. Then the extension being non-hex makes namespace collisions impossible. Then dirs don't need an extension.

Keys can also be raw binary, which makes the above necessary anyway.

And the encoding can't be base58 if you care about HFS.

But if it doesn't look pretty in `find`, what's the purpose of preserving the key segments, and not just flattening it (treating the key as opaque binary)? To support faster `Query` prefixes, a feature that isn't used anyway?

For the record, I'm tempted to make the blockstore use a flat version and not this. It's just way simpler and robust, don't have to worry about all the edge cases.

jbenet, line 32: > The real fix is encoding every segment, e.g. hex. Then the extension being non-hex makes namespace collisions impossible. Then dirs don't need an extension.
> 
> Keys can also be raw binary, which makes the above necessary anyway.

Yeah that seems a good way to do it.

> And the encoding can't be base58 if you care about HFS.

Yep, https://github.com/ipfs/specs/tree/master/repo/fs-repo#blocks-with-an-fs-datastore <-- notes this. we dont have to use the approach listed there-- we can modify it if desired. Just want to keep the objects inspectable by users, like git.

> But if it doesn't look pretty in find, what's the purpose of preserving the key segments, and not just flattening it (treating the key as opaque binary)? To support faster Query prefixes, a feature that isn't used anyway?

Both Query prefixes and user readability.

>  I'm tempted to make the blockstore use a flat version and not this. It's just way simpler and robust, don't have to worry about all the edge cases.

it'll be way easier for us to implement an `s3-datastore` if we keep it in terms of datastore :/ (and even tier them to treat the `fs-datastore` as a local cache). **this already important, as we need some ipfs nodes that store blocks in s3**

---

Actually how about this. This fs-datastore is tuned for user viewing-- i.e. the keys are the same -- which is not the best for the go-ipfs repo. Maybe we can make a different fs-datastore that encodes all the keys with hex, etc.

tv42, line 32: The flat version would still be just a Datastore, it'd just use the filesystem differently than this package. S3 plans are not affected by that.

Summarizing: we leave this package as is, with some prominent caveats that it should only be used with human-friendly, trusted keys, and I make a new package alongside. That sounds good.

jbenet, line 32: Perfect!

tv42:
Closing in favor of https://github.com/jbenet/go-datastore/pull/16

PR #21, opened 2015-05-28
switch uuid implementation

cryptix:
motivation: code.google.com is going away and satori's seemd sufficient and not overly complex


jbenet:
LGTM

PR #22, opened 2015-06-25
add in support for batched writes

whyrusleeping:
This implements a transaction object to allow for puts and deletes to be combined into a single operation. On certain backends this provides very significant performance improvements. Testing flatfs on my 700MB/s read/write SSD with 20MB worth of 100k writes showed a nearly 3x performance improvement (operation went from taking 1.6seconds to about 0.6 seconds)

### DIFF
--- a/basic_ds.go
+++ b/basic_ds.go
@@ -63,6 +63,10 @@ func (d *MapDatastore) Query(q dsq.Query) (dsq.Results, error) {
 	return r, nil
 }
 
+func (d *MapDatastore) StartBatchOp() Transaction {
+	return NewBasicTransaction(d)
+}
+
 // NullDatastore stores nothing, but conforms to the API.
 // Useful to test with.
 type NullDatastore struct {
@@ -96,6 +100,10 @@ func (d *NullDatastore) Delete(key Key) (err error) {
 // Query implements Datastore.Query
 func (d *NullDatastore) Query(q dsq.Query) (dsq.Results, error) {
 	return dsq.ResultsWithEntries(q, nil), nil
+}
+
+func (d *NullDatastore) StartBatchOp() Transaction {
+	return NewBasicTransaction(d)
 }
 
 // LogDatastore logs all accesses through the datastore.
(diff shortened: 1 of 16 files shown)


whyrusleeping:
I also need to implement the interface for the other datastores here, but for now. some CR on whats been done.

basic_ds.go
```diff
@@ -63,6 +63,10 @@ func (d *MapDatastore) Query(q dsq.Query) (dsq.Results, error) {
 	return r, nil
 }
 
+func (d *MapDatastore) StartBatchOp() Transaction {
```
jbenet: how about just `Batch()`

datastore.go
```diff
@@ -104,3 +107,11 @@ func GetBackedHas(ds Datastore, key Key) (bool, error) {
 		return false, err
 	}
 }
+
+type Transaction interface {
```
jbenet: let's call this a `Batch` because `Transaction` implies transactional semantics, with atomicity and potentially rollback guarantees. None of that here. We could add a:

``` go
type Batch interface {
    IsTransactional() bool
}
```

to check whether a batch is transactional or not.

transaction.go
```diff
@@ -0,0 +1,44 @@
...
+// basicTransaction implements the transaction interface for datastores who do
+// not have any sort of underlying transactional support
+type basicTransaction struct {
+	puts    map[Key]interface{}
+	deletes map[Key]struct{}
+
+	target Datastore
+}
+
+func NewBasicTransaction(ds Datastore) Transaction {
```
jbenet: `NewBasicBatch`

transaction.go
```diff
@@ -0,0 +1,44 @@
...
+	return &basicTransaction{
+		puts:    make(map[Key]interface{}),
+		deletes: make(map[Key]struct{}),
+		target:  ds,
+	}
+}
```
jbenet, line 18: add

``` go
func (bt *basicBatch) IsTransactional() bool { return false }
```

whyrusleeping, line 18: this does nothing for us, I'll just rename to batch.

jbenet, line 18: it does, it makes it clear that this is not transactional and is not atomic and people should not expect it to be.

jbenet, line 18: (and it allows future datastores that are to signal it)

jbenet:
i think we should not grow the interface with this batch thing-- it already broke tons of tools i have out there.

I'm going to revert. we should use an interface like:

``` go
type Batcher interface {
  Batch() Batch
}

type BatcherDatastore interface {
  Datastore
  Batcher
}
```

PR #23, opened 2015-06-27
Revert "add in support for batched writes"

jbenet:
Reverts jbenet/go-datastore#22

see https://github.com/jbenet/go-datastore/pull/22#issuecomment-116020211


jbenet:
cc @whyrusleeping

PR #25, opened 2015-07-01
Batching

whyrusleeping:
Restarting this PR


keytransform/interface.go
```diff
@@ -16,6 +16,8 @@ type KeyTransform interface {
 type Datastore interface {
 	ds.Shim
 	KeyTransform
+
+	Batch() (ds.Batch, error)
```
jbenet, line 20: maybe `BatchingDatastore` instead

sync/sync.go
```diff
@@ -63,3 +63,14 @@ func (d *MutexDatastore) Query(q dsq.Query) (dsq.Results, error) {
...
+
+func (d *MutexDatastore) Batch() (ds.Batch, error) {
+	d.RLock()
+	defer d.RUnlock()
+	bds, ok := d.child.(ds.BatchingDatastore)
+	if !ok {
+		return nil, ds.ErrBatchUnsupported
+	}
+
+	return bds.Batch()
```
jbenet: this is incorrect. need to make a special batch that locks+unlocks `d.Lock()` on `batch.{Put,Delete}`

mount/mount.go
```diff
@@ -114,3 +114,62 @@ func (d *Datastore) Query(q query.Query) (query.Results, error) {
...
+}
+
+func (mt *mountBatch) Commit() error {
+	for _, t := range mt.mounts {
+		err := t.Commit()
+		if err != nil {
+			return err
+		}
+	}
+	return nil
```
jbenet, line 174: want to clear the batches at the end. +Lock.

mount/mount.go
```diff
@@ -114,3 +114,62 @@ func (d *Datastore) Query(q query.Query) (query.Results, error) {
...
+			return nil, datastore.NewKey(""), datastore.ErrBatchUnsupported
+		}
+		var err error
+		t, err = bds.Batch()
+		if err != nil {
+			return nil, datastore.NewKey(""), err
+		}
+		mt.mounts[loc.String()] = t
+	}
+	return t, rest, nil
```
jbenet, line 146: this lookup should happen with a lock on the `mountBatch` (sync access to `mt.mounts`)

mount/mount.go
```diff
@@ -114,3 +114,62 @@ func (d *Datastore) Query(q query.Query) (query.Results, error) {
 	r = query.ResultsReplaceQuery(r, q)
 	return r, nil
 }
+
+type mountBatch struct {
+	mounts map[string]datastore.Batch
```
jbenet, line 119: needs a lock

measure/measure.go
```diff
@@ -148,6 +148,81 @@ func (m *measure) Query(q query.Query) (query.Results, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &measuredBatch{
+		putts: pb,
+		delts: db,
+
+		m: m,
+	}, nil
```
jbenet, line 181: i dont understand why `putts` and `delts` need to be different batches?

measure/measure.go
```diff
@@ -148,6 +148,81 @@ func (m *measure) Query(q query.Query) (query.Results, error) {
...
+}
+
+func (mt *measuredBatch) Put(key datastore.Key, val interface{}) error {
+	mt.puts++
+	return mt.putts.Put(key, val)
+}
+
+func (mt *measuredBatch) Delete(key datastore.Key) error {
+	mt.deletes++
```
jbenet, line 195: sync

measure/measure.go
```diff
@@ -148,6 +148,81 @@ func (m *measure) Query(q query.Query) (query.Results, error) {
...
+		took := int(time.Now().Sub(before)/time.Microsecond) / mt.deletes
+		mt.m.deleteNum.AddN(uint64(mt.deletes))
+		for i := 0; i < mt.deletes; i++ {
+			mt.m.deleteLatency.RecordValue(int64(took))
+		}
+		if err != nil {
+			mt.m.deleteErr.Add()
+			return err
+		}
+	}
```
jbenet, line 203: all these ops should be a function so that we dont screw up in once place and not all.

keytransform/interface.go
whyrusleeping, line 20: change the entire interface name? or have a new interface for batching?

mount/mount.go
whyrusleeping, line 174: why does this need to be locked? Also, batches are not supposed to be reused, what would the point of clearing them be?

mount/mount.go
whyrusleeping, line 146: I didnt think datastores were inherently threadsafe, thus the need for the mutexDatastore

measure/measure.go
whyrusleeping, line 181: discussed in person, how should the measurements work if we combine them into one?

measure/measure.go
whyrusleeping, line 195: again, i didnt think that datastores were supposed to be threadsafe

mount/mount.go
jbenet, line 146: i guess there are two sorts of "threadsafe"-- there's (a) concurrent access is not guaranteed to be consistent (i.e. two writes to the same key may interleave (like a fs), and (b) concurrent access will panic?

mount/mount.go
jbenet, line 174: > Also, batches are not supposed to be reused, what would the point of clearing them be?

ah, ok. that makes sense.

measure/measure.go
jbenet, line 181: yeah guess leave it as is

keytransform/interface.go
jbenet, line 20: > change the entire interface name? or have a new interface for batching?

i just meant maybe embed the interface instead of the method name, but embedding the interface will probably clash with `ds.Shim` so nvm.

whyrusleeping:
@jbenet ping

flatfs/flatfs.go
```diff
@@ -86,6 +82,17 @@ func (fs *Datastore) makePrefixDir(dir string) error {
 	return nil
 }
 
+func (fs *Datastore) makePrefixDirNoSync(dir string) error {
+	if err := os.Mkdir(dir, 0777); err != nil {
+		// EEXIST is safe to ignore here, that just means the prefix
+		// directory already existed.
```
jbenet, line 88: may need to do `Mkdirp` on the dirname? because the prefix may be longer? (maybe not)

jbenet:
@whyrusleeping code LGTM but... there's like no test :cry: :cry:

jbenet:
merging things without tests = recipe for painful times ahead

whyrusleeping:
image

jbenet:
@whyrusleeping

flatfs/flatfs.go
whyrusleeping, line 88: hrm, this was @tv42's code, i just broke the function into two separate calls

whyrusleeping:
@jbenet (tests appeared magically?)

flatfs/flatfs.go
tv42, line 88: What's `Mkdirp`?

jbenet, line 88: oh i meant http://golang.org/pkg/os/#MkdirAll but it's not _needed_

jbenet:
@whyrusleeping remaining:
- [x] test things are not there before
- [x] generic batching test

flatfs/flatfs.go
tv42, line 88: Whenever you see me not using MkdirAll, it's 99% safe to assume it's intentional. The parent dir should exist by the time that code runs; if not, I actually want to see an error. This also makes typos near the beginning of the path less likely to create weird shadow hierarchies. The Init function that creates the parent is https://github.com/ipfs/go-ipfs/blob/b1bb16f0f291d66b5902b8a6cefc2eac204a31e2/repo/fsrepo/fsrepo.go#L263

jbenet, line 88: :+1: sgtm

whyrusleeping:
@jbenet 7b9bbbc

PR #27, opened 2015-07-17
add threadsafe mount datastore and small fixups on keytransform

whyrusleeping:
Some other changes i realized i needed once i started diving into getting this integrated into go-ipfs.

### DIFF
--- a/keytransform/interface.go
+++ b/keytransform/interface.go
@@ -16,8 +16,6 @@ type KeyTransform interface {
 type Datastore interface {
 	ds.Shim
 	KeyTransform
-
-	Batch() (ds.Batch, error)
 }
 
 // Wrap wraps a given datastore with a KeyTransform function.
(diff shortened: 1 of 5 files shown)


keytransform/interface.go
```diff
@@ -16,12 +16,14 @@ type KeyTransform interface {
 type Datastore interface {
 	ds.Shim
 	KeyTransform
+
+	Batch() (ds.Batch, error)
```
jbenet: KeyTransform _requires_ Batch? returning *ktds returns a `ds.Batching` already...

jbenet:
LGTM otherwise

keytransform/interface.go
whyrusleeping: oh, this change is probably not necessary now that i've changed the constructor to return `*ktds`

PR #29, opened 2015-09-10
allow put to try again if it fails due to too many open files

whyrusleeping:
A small workaround to try and mitigate the 'too many open file descriptors' issues we are seeing.


whyrusleeping:
One issue i have with this is that we dont get to see when these temporary failures occur. It might be okay, but i'm paranoid.

jbenet:
Log them?

(we should make the `util.Logger` thing its own package or something, so go-datastore and go-commands can import it)

whyrusleeping:
> (we should make the util.Logger thing its own package or something, so go-datastore and go-commands can import it)

I'll do that, its a good idea.

jbenet:
@whyrusleeping squash

whyrusleeping:
squashed

PR #30, opened 2015-09-27
Add sync flag to flatfs

rht:
(should only be enabled for the paranoid, cc: @tv42 and https://github.com/ipfs/archives/issues/20)

### DIFF
--- a/flatfs/flatfs.go
+++ b/flatfs/flatfs.go
@@ -34,11 +34,16 @@ type Datastore struct {
 	path string
 	// length of the dir splay prefix, in bytes of hex digits
 	hexPrefixLen int
+	sync         bool
 }
 
 var _ datastore.Datastore = (*Datastore)(nil)
 
 func New(path string, prefixLen int) (*Datastore, error) {
+	return NewMaybeSync(path, prefixLen, true)
+}
+
+func NewMaybeSync(path string, prefixLen int, sync bool) (*Datastore, error) {
 	if prefixLen <= 0 || prefixLen > maxPrefixLen {
 		return nil, ErrBadPrefixLen
 	}
@@ -46,6 +51,7 @@ func New(path string, prefixLen int) (*Datastore, error) {
 		path: path,
 		// convert from binary bytes to bytes of hex encoding
 		hexPrefixLen: prefixLen * hex.EncodedLen(1),
+		sync:         sync,
 	}
 	return fs, nil
 }
@@ -77,12 +83,12 @@ func (fs *Datastore) makePrefixDir(dir string) error {
 		return err
 	}
 
-	// In theory, if we create a new prefix dir and add a file to
-	// it, the creation of the prefix dir itself might not be
-	// durable yet. Sync the root dir after a successful mkdir of
-	// a prefix dir, just to be paranoid.
-	if err := syncDir(fs.path); err != nil {
-		return err
+	if fs.sync {
+		// In theory, if we create a new prefix dir and add a file to
+		// it, the creation of the prefix dir itself might not be
+		// durable yet. Sync the root dir after a successful mkdir of
+		// a prefix dir, just to be paranoid.
+		return syncDir(fs.path)
 	}
 	return nil
 }
@@ -149,8 +155,10 @@ func (fs *Datastore) doPut(key datastore.Key, val []byte) error {
 	if _, err := tmp.Write(val); err != nil {
 		return err
 	}
-	if err := tmp.Sync(); err != nil {
-		return err
+	if fs.sync {
+		if err := tmp.Sync(); err != nil {
+			return err
+		}
 	}
 	if err := tmp.Close(); err != nil {
 		return err
@@ -163,8 +171,8 @@ func (fs *Datastore) doPut(key datastore.Key, val []byte) error {
 	}
 	removed = true
 
-	if err := syncDir(dir); err != nil {
-		return err
+	if fs.sync {
+		return syncDir(dir)
 	}
 	return nil
 }
@@ -214,8 +222,10 @@ func (fs *Datastore) putMany(data map[datastore.Key]interface{}) error {
 	// Now we sync everything
 	// sync and close files
 	for fi, _ := range files {
-		if err := fi.Sync(); err != nil {
-			return err
+		if fs.sync {
+			if err := fi.Sync(); err != nil {
+				return err
+			}
 		}
 
 		if err := fi.Close(); err != nil {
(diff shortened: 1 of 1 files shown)


jbenet:
Ugh -- writes to disk should be durable...

@whyrusleeping didn't you find that the syncs weren't the bottleneck?

rht:
> Ugh -- writes to disk should be durable...

`cp` doesn't do this [1], and not `git` by default either [2].

Perf increases by an order magnitude (1.5s -> 0.13s) with 10 1.1MB files, and 6min -> 1.7min with 999 1MB files. Of course 0.13s is still much slower compared to git/cp.

Reproducible test script (or if someone already wrote a benchmark script for git vs cp vs rsync etc) coming soon like @ion1 did in https://github.com/ipfs/go-ipfs/issues/1750.

[1] https://github.com/coreutils/coreutils/search?utf8=%E2%9C%93&q=fsync
[2] https://github.com/ipfs/archives/issues/20#issuecomment-143511009 copied here for convenience

> Also, found this http://git.kernel.org/cgit/git/git.git/tree/Documentation/config.txt#n693:
> 
> > This is a total waste of time and effort on a filesystem that orders data writes properly, but can be useful for filesystems that do not use journalling (traditional UNIX filesystems) or that only journal metadata and not file contents (OS X's HFS+, or Linux ext3 with "data=writeback").

jbenet:
Hm, how do @tv42 @whyrusleeping @cryptix @chriscool feel about this?

---

As far as IPFS is concerned, maybe we can make this a config setting, or even be able to do it per-command (like `ipfs add --no-sync` type of thing).

tv42:
I feel comparing `cp` to something providing services to other hosts is apples to oranges wrt crash safety.

The git config option seems to be only about loose objects; there still is an `fsync_or_die` in e.g. pack creation -- that is, there is an fsync for pushed data. Git is fast because of a series of smart design decisions, not because it's unsafe.

Would like to hear what @whyrusleeping has to say about his experiments.

jbenet:
> I feel comparing cp to something providing services to other hosts is apples to oranges wrt crash safety.

:+1:

rht:
> I feel comparing cp to something providing services to other hosts is apples to oranges wrt crash safety.

`ipfs add` is used under several conditions, some of which intersects with `cp` (offline add where the blocks being added don't have to be streamed live), while some others (distributed db on ipfs) don't.

Ok the fsync flag in git is only used in loose object writes.
Maybe fsync is not git's bottleneck (since its perf ~ cp perf), but it is for flatfs.
edit: but git is fast even _with_ fsync.

chriscool:
When writing many files or creating many directories, I think it is probably overkill to fsync after each write. It might be a good idea to sync when all the writes are done though.

chriscool:
I mean that it is impossible anyway to prevent crashes from happening while writing files (or between writing files), but it could be a good idea to sync after writing all the stuff and before advertising it on the network, to make sure that everything advertised has been writen on disk at one point. Though it is of course impossible to ensure that a disk crash will not wipe out the content anyway.

jbenet:
@chriscool yeah, i believe that's what @whyrusleeping's attempt did previously -- fsync at the and of all writes. would that be good enough for you rht?

also if we had a write-ahead log style arena-storage repo, it might work well.

rht:
The current implementation is _already_ to fsync after all files have been written to tmpfiles, then `mv` to 'blocks/'.
This PR is meant to make these explicit fsync calls optional.

I have no idea what magic git uses, but I'd opt for no-sync as default, and only do `ipfs add --sync` for ipdb. My reason: the former case ("offline" batch add) is more common today.

Still relevant today: https://web.archive.org/web/20150315020954/http://thunk.org/tytso/blog/2009/03/15/dont-fear-the-fsync/ ?

jbenet:
I think right now, if we add the option to disable fsyncs, it should be an explicit opt in.

I think later, we can move to mmapped files and perhaps a journal. That will likely have a different sync model with different knobs to turn.

flatfs/flatfs.go
```diff
@@ -34,11 +34,12 @@ type Datastore struct {
 	path string
 	// length of the dir splay prefix, in bytes of hex digits
 	hexPrefixLen int
+	sync         bool
 }
 
 var _ datastore.Datastore = (*Datastore)(nil)
 
-func New(path string, prefixLen int) (*Datastore, error) {
+func New(path string, prefixLen int, sync bool) (*Datastore, error) {
```
jbenet: perhaps we should keep `New` as is, setting sync to `true` and just add another constructor:

```
NewSync(...) 
```

with the bool flag option. this prevents changing the interface on whoever depends on this code

rht:
(assuming sync is disabled.
Still with the 10 1.1MB files test, the bulk of the 0.13s comes from merkledag pb Marshal, in particular the line https://github.com/ipfs/go-ipfs/blob/master/merkledag/pb/merkledag.pb.go#L557 (~100ms).
One option is to use unsafe_marshaler_all in the protobuf generation.
The doc says, https://github.com/gogo/protobuf/blob/master/plugin/marshalto/marshalto.go#L44, but currently checking how fast it could be.
e.g. syscall.Read in zsyscall_linux_*.go opts to use unsafe pointer.
)

> we can move to mmapped files and perhaps a journal.

Mind to elaborate?
For the latter, what if the host fs is already journaled?
Here is another compromise: run this once after all the files have been written to tmpfiles https://golang.org/pkg/syscall/#Sync.

cryptix:
If this is about `ipfs add` performance and not a consistency issue, it feels like a band-aid patch to me.

rht:
@cryptix the per-file sync is the most significant bottleneck. Even if add perf issue were to be addressed later, it will have to go down first.

chriscool:
>  Here is another compromise: run this once after all the files have been written to tmpfiles
> https://golang.org/pkg/syscall/#Sync.

Yeah, this might be worth trying. It should just call: https://www.cl.cam.ac.uk/cgi-bin/manpage?2+sync and might be more efficient than many calls to fsync

rht:
> (This still does not guarantee data integrity: modern disks have large caches.)

With `syscall.Sync()`: 1.3s -> ~0.9s.

chriscool:
@rht did you try with 10 1.1MB files? What about with 999 1MB files?

jbenet:
It would be very useful to have good benchmarks and graphs of all this, across disk types, host fses, and OSes. Maybe we can include this in our thought re golang/build

—

rht:
10 1MB files (actually it is 1): 1.3s -> 0.9s.
999 1MB files:  6 min -> 1.7.

(making the benchmark. should including adding updated files as well)

tv42:
Does that calling code use whyrusleeping's batching feature?

rht:
Iirc the batching is used in the `Commit()` call, since flatfsBatch uses `putMany`, then yes.
I modified `doPut` and `putMany` to use `syscall.Read()`.

(but then again, fsync is required for ipdb, but only for ipdb use case)

rht:
What is the decision on the `--no-sync` flag?
Benchmark is shown in #31, and it hasn't changed much in the latest master https://github.com/ipfs/go-ipfs/commit/b508c23b7e40837cfbaaef58ce5a64253589808d.

Notice that even sqlite has http://www.sqlite.org/pragma.html#pragma_synchronous.

Here is the rationale behind the rare usage of fsync in git: https://lwn.net/Articles/326505/, https://plus.google.com/+JonathanCorbet/posts/JBxiKPe3VXa (+comments).

rht:
(duplicate of https://github.com/jbenet/go-datastore/pull/32)

PR #31, opened 2015-10-03
Replace fsync with global sync call

rht:


rht:
On ext4

fsync:
outdata

sync:
outdata_sync

rht:
(benchmark script: https://ipfs.io/ipfs/Qma6m4ALV6VHVVijpVc8eRsjtGYK4LBLHUMyn6ZfkJnf7s)

rht:
nosync:
outdata

jbenet:
@rht thanks for these. However, i'm not sure how these scale with lots of other disk activity and multiple concurrent `sync` calls. I suspect hammering at ipfs with thousands of requests would show that fsync is better, but i haven't run the test. cc @whyrusleeping @tv42 

i still think no-sync should be an option, (sync by default). ideally we could do it on a per-command basis on ipfs. so one could `ipfs add --no-sync` or something.

rht:
In some filesystems, per-file `fsync` is already short-circuited to global `sync`. This PR simply makes it explicit that it is the default for any host fs.

Why would fsync be faster for the case of "lots of other disk activity and multiple concurrent sync calls"?

rht:
Benchmark already shows otherwise (sync vs fsync).

But now that there are two options (default sync for safe operation and nosync for fast one), this middle ground is no longer necessary.

PR #32, opened 2015-11-09
add option to disable syncing

whyrusleeping:
gives flatfs a way to be less safe, but wayyyyy faster.

### DIFF
--- a/flatfs/flatfs.go
+++ b/flatfs/flatfs.go
@@ -34,11 +34,14 @@ type Datastore struct {
 	path string
 	// length of the dir splay prefix, in bytes of hex digits
 	hexPrefixLen int
+
+	// sychronize all writes and directory changes for added safety
+	sync bool
 }
 
 var _ datastore.Datastore = (*Datastore)(nil)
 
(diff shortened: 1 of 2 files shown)


rht:
As with the interface, I thought the default `New` constructor syncs by default? ref: https://github.com/jbenet/go-datastore/pull/30#discussion_r40523430

jbenet:
i'd vote for syncing by default-- but i guess this at least breaks the code and forces users to look at the new interface.

if we want to not break code, i'd suggest:

``` go
func New(path string, prefixLen int) (*Datastore, error)
func NewSync(path string, prefixLen int, sync bool) (*Datastore, error)
```

rht:
As with compatibility breaking, the project that vendors and uses go-datastore _directly_ so far is go-ipfs (that I can find from the internet). And so this can be quickly mended.

note: all the specific implementations of the Datastore interface has one version of `New`/`NewDatastore`/`NewDs`. Some of them use option struct (just as struct fields have default value, and hence the defaults can be set), which is more future-proof. But this is unnecessary unless there will be more options to `flatfs.New` in the future.

PR #34, opened 2015-12-13
use filepath walk for flatfs query

whyrusleeping:
running a query on a large ipfs datastore takes forever (and uses a lot of ram), this should help, some.

### DIFF
--- a/flatfs/flatfs.go
+++ b/flatfs/flatfs.go
@@ -9,6 +9,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path"
+	"path/filepath"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


flatfs/flatfs.go
```diff
@@ -314,52 +315,33 @@ func (fs *Datastore) Query(q query.Query) (query.Results, error) {
...
-	for _, fi := range prefixes {
-		var err error
-		res, err = fs.enumerateKeys(fi, res)
-		if err != nil {
-			return nil, err
-		}
-	}
-	return query.ResultsWithEntries(q, res), nil
-}
+			if !info.Mode().IsRegular() || info.Name()[0] == '.' {
```
jbenet, line 323: `.` but no `..`?

jbenet:
this LGTM

flatfs/flatfs.go
whyrusleeping, line 323: I think the intent of this check was to ignore the random dotfiles that os'es (OSX) will randomly decide to put in folders. Other than that, i'm not certain.

jbenet, line 323: Oh name[0], we're good.

PR #35, opened 2015-12-27
fix locking in syncbatch

whyrusleeping:
If the batch returned by the underlying batching datastore isnt properly locked, calling commit on the  batch (even if it takes its lock) doesnt respect the lock in the MutexDatastore that we are using to protect the given datastore. This uses the lock on the MutexDatastore to do all locking for the child batch.

It uses it for all batch operations as some batches may write to the datastore outside of commit being called.


jbenet:
Good fix. LGTM

PR #37, opened 2016-02-19
Print formatted string using Errorf

rikonor:
Hi,

Currently `log.Error` is used with a formatted string, causing output of the form:

```
flatfs: too many open files, retrying in %dms0 flatfs.go:119
```

Switched `log.Error` with `log.Errorf`.


kevina:
@whyrusleeping this LGTM

Is there a way we can get it to trigger the integration tests?

Kubuxu:
Also LGTM.

PR #41, opened 2016-06-28
omit slash from created dskeys

whyrusleeping:
re: @kevina's comments after my last PR

### DIFF
--- a/flatfs/flatfs.go
+++ b/flatfs/flatfs.go
@@ -56,9 +56,10 @@ func New(path string, prefixLen int, sync bool) (*Datastore, error) {
 var padding = strings.Repeat("_", maxPrefixLen)
 
 func (fs *Datastore) encode(key datastore.Key) (dir, file string) {
-	prefix := (key.String() + padding)[:fs.prefixLen]
+	noslash := key.String()[1:]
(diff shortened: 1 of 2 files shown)


flatfs/flatfs.go
```diff
@@ -56,7 +56,7 @@ func New(path string, prefixLen int, sync bool) (*Datastore, error) {
 var padding = strings.Repeat("_", maxPrefixLen)
 
 func (fs *Datastore) encode(key datastore.Key) (dir, file string) {
-	prefix := (key.String() + padding)[:fs.prefixLen]
+	prefix := (key.String() + padding)[1:fs.prefixLen]
 	dir = path.Join(fs.path, prefix)
 	file = path.Join(dir, key.String()+extension)
```
kevina: This should be:

```
noslash := key.String()[1:]
prefix := (noslash + padding)[:fs.prefixLen]
dir = path.Join(fs.path, prefix)
file = path.Join(dir, noslash+extension)
```

as go slices are `[start:stop]` not `[start:len]`.

kevina:
LGTM.

Kubuxu:
LGTM.

PR #48, opened 2016-08-25
Feat/extract flatfs

whyrusleeping:
choo choo! the modularization train is moving!


Kubuxu:
Where have you moved it to?

whyrusleeping:
https://github.com/ipfs/go-ds-flatfs

Issue #47, opened 2016-08-17
split out subpackages into separate packages


whyrusleeping:
Updating portions of this codebase would be much easier if some of the more used subcomponents were extracted, such as flatfs and leveldb.

Restricting this package to an interface definition and some small 'stock' implementations would also greatly improve the dependency tree of any application wanting to make use of this codebase

jbenet:
Sgtm!

whyrusleeping:
I've extracted a lot of the subrepos. Anything that had external dependencies now lives outside of this repo. I think for now, we can close this. If packages in this repo end up getting changes made to them more frequently the we should consider moving them out as well.

For now, lets close this

PR #49, opened 2016-08-26
extract more subpackages

whyrusleeping:
I extracted every subpackage that required an external import. The others can stay for now (especially since they arent frequently used)


whyrusleeping:
mostly resolves #47

Kubuxu:
LGTM

PR #52, opened 2016-11-14
Fix channel buffer size in namespace datastore to use KeysOnlyBufSize

kevina:


kevina:
See ipfs/go-ipfs#3376.

I better way would be to use something like `make(chan dsq.Result, cap(qr.Next()))` but for some reason that cap(qr.Next()) is returning 0 when it should be returning 128 (i.e. dsq.KeysOnlyBufSize) so I manually set the buffer size based on it it was a keys-only query.

kevina:
Note the test failure is due to the use of go 1.6.3 that no longer appears to be supported with the latest version of go-datastore.

Kubuxu:
Can you update the travis file to use 1.7.3 and tip.

kevina:
@Kubuxu 

> Can you update the travis file to use 1.7.3 and tip.

Done.

whyrusleeping:
Sweet, LGTM. I think the biggest problem here is that we're trying to use a channel as an iterator and thats just inherently slow. We can look at switching up interfaces later, but the changes being made so far look very promising

PR #53, opened 2016-11-14
add non-channel query iterator method

whyrusleeping:
Using channels as query iterators is quite inefficient. A naive implementation of the leveldb query results that avoids using a channel improves perf by close to 20%.


kevina:
This could work, but it might complicate composing of datastores such as the namespace datastore.

PR #55, opened 2016-11-16
Make sure buffer size of wrapper datastores match underlying datastore.

kevina:
This is the way I wanted to fix #52.  It turned out there was yet another channel I didn't notice originally.  With this fix the buffer size of all the channels is now 128 for `blockstore.AllKeysChan()`.

This decreased the time to retrieve 100,000 small (1000 byte) blocks (see ipfs/go-ipfs#3376) from around 340 ms to 245 ms.


whyrusleeping:
LGTM

PR #56, opened 2016-11-16
Add RawKey function and make use of it.

kevina:
Avoids some unnecessary creation of strings.

This decreased the time to retrieve 100,000 small (1000 byte) blocks (see ipfs/go-ipfs#3376) from around 245 ms to 205 ms.

### DIFF
--- a/namespace/namespace.go
+++ b/namespace/namespace.go
@@ -2,7 +2,6 @@ package namespace
 
 import (
 	"fmt"
-	"strings"
 
 	ds "github.com/ipfs/go-datastore"
 	ktds "github.com/ipfs/go-datastore/keytransform"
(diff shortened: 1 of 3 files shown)


kevina:
It looks like I broke something.  Will fix.

kevina:
Should be fixed now.

kevina:
@whyrusleeping should be good to go

namespace/namespace.go
```diff
@@ -29,8 +28,8 @@ func PrefixTransform(prefix ds.Key) ktds.KeyTransform {
 				panic("expected prefix not found")
 			}
 
-			s := strings.TrimPrefix(k.String(), prefix.String())
-			return ds.NewKey(s)
+			s := k.String()[len(prefix.String()):]
```
whyrusleeping, line 31: You should make sure to do a bounds check here.

kevina: I could but then what, panic.  Also it is redundant because if prefix.IsAncestorOf passes then there is no way for this to be out of bounds.

whyrusleeping: Ah, i see. I didnt realize we already had that expectation here.

whyrusleeping:
LGTM, thanks!`

PR #57, opened 2016-11-17
namespace: correctly handle case when prefix is empty

kevina:
Bug fix for #56.


whyrusleeping (approved):
Looks like a solid optimization to me

PR #58, opened 2016-11-17
Avoid creating unnecessary strings when key is already in correct format

kevina:
I have not performed any tests but this should help just about everywhere.


whyrusleeping (approved):
mmm, yeah. those little string operations are death by 1000 cuts

PR #63, opened 2017-03-14
test: fix query test

Kubuxu:
String keys are prefixed with /, which was causing failures as it
expects key without the slash.

Checked and it now runs on go-ds-flatfs


whyrusleeping (approved):
LGTM

PR #65, opened 2017-06-16
Log more in LogDatastore

magik6k:
* `logds.Query` logs more info
* `logds.Batch` wraps the batch and logs operations on it


whyrusleeping (approved):
LGTM

PR #66, opened 2017-06-28
Implemented cross-ds query in mount datastore

magik6k:
* mountds.Query now works across datastores
* Mounts are sorted to mitigate `lookup` dependency on ordered keys
* `syncmount` tests now actually test syncmount DS

### DIFF
--- a/mount/mount.go
+++ b/mount/mount.go
@@ -9,7 +9,6 @@ import (
 	"strings"
 
 	"github.com/ipfs/go-datastore"
-	"github.com/ipfs/go-datastore/keytransform"
 	"github.com/ipfs/go-datastore/query"
 )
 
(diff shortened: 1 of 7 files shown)


mount/mount.go
```diff
@@ -21,12 +21,27 @@ type Mount struct {
 	Datastore datastore.Datastore
 }
 
+type MountSlice []Mount
```
whyrusleeping: you *could* use the new `sort.Slice` api, which saves you a good bit of code and a type. Though either way is fine with me

mount/mount.go
```diff
@@ -47,6 +62,42 @@ func (d *Datastore) lookup(key datastore.Key) (ds datastore.Datastore, mountpoin
...
+// Matching: /ao/e
+//
+// /          B /ao/e
+// /a/        not matching
+// /ao/       B /e
+// /ao/e/     A /
+// /ao/e/uh/  A /
+// /aoe/      not matching
+func (d *Datastore) lookupAll(key datastore.Key) (ds []datastore.Datastore, mountpoint, rest []datastore.Key) {
+	ds = []datastore.Datastore{}
```
whyrusleeping: you can append to `nil`, no need to set values on these here

whyrusleeping:
I would love to be able to reuse code between mount and syncmount, but we can probably add that as a todo for later improvements

magik6k:
I'd make it an issue for now

whyrusleeping (approved):
LGTM!

PR #67, opened 2017-07-03
Fix subquery closing in mount, improve namespace.Query

magik6k:
Tested on go-ipfs, this seems to fix related issues.

### DIFF
--- a/mount/mount.go
+++ b/mount/mount.go
@@ -126,9 +126,7 @@ func (d *Datastore) Query(q query.Query) (query.Results, error) {
 
 	// current itorator state
 	var res query.Results
-	var ds datastore.Datastore
 	var mount datastore.Key
-	var rest datastore.Key
 	i := 0
 
 	return query.ResultsFromIterator(q, query.Iterator{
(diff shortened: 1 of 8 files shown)


mount/mount.go
```diff
@@ -137,19 +135,19 @@ func (d *Datastore) Query(q query.Query) (query.Results, error) {
 			var more bool
 
 			for try := true; try; try = len(dses) > i {
-				if ds == nil {
+				if res == nil {
```
whyrusleeping, line 138: just a personal preference that i should have caught in the previous PR, but we should probably add the `res != nil` check to the for-loop condition, since it essentially says "loop until res != nil or try is false.

whyrusleeping (approved):
one style nitpick that doesnt need to be changed now (was from a previous PR) and then make sure the tests successfuly reproduce the panics we're fixing. Aside from that, LGTM

mount/mount.go
magik6k: Note that this if doesn't contain the whole loop

whyrusleeping: oooooh, i see. Githubs review UI hid that from me. Carry on

whyrusleeping:
@magik6k any idea why the namespace tests would be failing here?

magik6k:
Looks like MapDatastore is not ordered

PR #69, opened 2017-10-16
Implement json.Marshaler and json.Unmarshaler on datastore.Key

b5:
Not sure weather this will be acceptable, but I'd love to be able to marshal / unmarshal datastore.Key values to and from JSON strings for clearer struct definitions that work with `datastore.Key`'s. We're doing all kinds of off-label stuff with the datastore interface, and this would be a huge help, I could also see potential upsides in IPLD land.

That being said, I can understand why marshaling to and from datastore.Key's may _not_ be a desired behavior, and would be more than happy to close this PR unmerged in that case.

Assuming you're ok with the idea of keys going to and from JSON, there are three options to get `datastore.Key`'s to do JSON:

1. **`type Key struct{ string }` -> `type Key string`**
 _I'm guessing you have a good reason for key's being a struct type, so leaving that alone_
2. **`import encoding/json`, implement `json.Un/Marshaler` with string types** 
_ew gross_
3. **implement a bare bones `json.Un/Marshaler`**
  _what this pr proposes_

I think the third option makes the most sense, it has the downside that if people put horribly malformed stuff in their json it'll just coerce that field to a string and not complain, which might be ugly, but seems both unlikely and not unlike the behavior of other higher-level languages. The upside is no added dependencies for such a foundational package.

### DIFF
--- a/key.go
+++ b/key.go
@@ -1,6 +1,7 @@
 package datastore
 
 import (
+	"errors"
 	"path"
 	"strings"
 
@@ -230,6 +231,23 @@ func (k Key) IsDescendantOf(other Key) bool {
 // IsTopLevel returns whether this key has only one namespace.
 func (k Key) IsTopLevel() bool {
 	return len(k.List()) == 1
+}
+
+// MarshalJSON implements the json.Marshaler interface,
+// keys are represented as JSON strings
+func (k Key) MarshalJSON() ([]byte, error) {
+	return []byte(`"` + k.String() + `"`), nil
+}
+
+// MarshalJSON implements the json.Unmarshaler interface,
+// keys will parse any value specified as a key to a string
+func (k *Key) UnmarshalJSON(data []byte) error {
+	if len(data) < 2 {
+		k.Clean()
+		return errors.New("too short to unmarshal key to json string")
+	}
+	*k = NewKey(string(data[1 : len(data)-1]))
+	return nil
 }
 
 // RandomKey returns a randomly (uuid) generated key.
--- a/key.go
+++ b/key.go
@@ -1,7 +1,7 @@
 package datastore
 
 import (
-	"errors"
+	"encoding/json"
 	"path"
 	"strings"
 
@@ -236,17 +236,17 @@ func (k Key) IsTopLevel() bool {
 // MarshalJSON implements the json.Marshaler interface,
 // keys are represented as JSON strings
 func (k Key) MarshalJSON() ([]byte, error) {
-	return []byte(`"` + k.String() + `"`), nil
+	return json.Marshal(k.String())
 }
 
 // MarshalJSON implements the json.Unmarshaler interface,
 // keys will parse any value specified as a key to a string
 func (k *Key) UnmarshalJSON(data []byte) error {
-	if len(data) < 2 {
-		k.Clean()
-		return errors.New("too short to unmarshal key to json string")
+	var key string
+	if err := json.Unmarshal(data, &key); err != nil {
+		return err
 	}
-	*k = NewKey(string(data[1 : len(data)-1]))
+	*k = NewKey(key)
 	return nil
 }
 
--- a/key_test.go
+++ b/key_test.go
@@ -163,6 +163,7 @@ func TestKeyMarshalJSON(t *testing.T) {
 		err  string
 	}{
 		{NewKey("/a/b/c"), []byte("\"/a/b/c\""), ""},
+		{NewKey("/shouldescapekey\"/with/quote"), []byte("\"/shouldescapekey\"/with/quote\""), ""},
 	}
 
 	for i, c := range cases {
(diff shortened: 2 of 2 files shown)


key.go
```diff
@@ -232,6 +233,23 @@ func (k Key) IsTopLevel() bool {
 	return len(k.List()) == 1
 }
 
+// MarshalJSON implements the json.Marshaler interface,
+// keys are represented as JSON strings
+func (k Key) MarshalJSON() ([]byte, error) {
+	return []byte(`"` + k.String() + `"`), nil
```
whyrusleeping: You might want to do some string escaping here, otherwise some keys might produce broken json

b5: very good point, will add

magik6k (commented):
I'm not exactly sure if we want to use json.Marshaler, there is a possibility that we well be moving to refmt - https://github.com/polydawn/refmt

whyrusleeping:
@magik6k that doesnt conflict with us adding these methods here. Most people will still use the standard json library, and this helps for that.

b5:
Slammed out a quick addition the the Marshal test, but will hold up on adding cases / changing code for the minute.

@magik6k, after checking out refmt (which looks fantastic), I think the two would play nicely together. This change doesn't import the json package (just interface satisfaction), and refmt/json doesn't check for the `json.Un/Marshaler` interface, so these funcs will never be called by that package.

b5:
lol apologies for repeating @whyrusleeping on a 10sec delay ;)

whyrusleeping:
@b5 it looks like the tests you wrote here are failing.

b5:
ah yes, sorry [this test import](https://github.com/ipfs/go-datastore/blob/dd51b0eacf47a84dc9965f6b9a7b4581d8f41284/key_test.go#L10) was tripping up local tests

b5:
@whyrusleeping, anything more you'd like to see from this PR?

key.go
whyrusleeping: has this been addressed?

key.go
```diff
@@ -232,6 +233,23 @@ func (k Key) IsTopLevel() bool {
...
+}
+
+// MarshalJSON implements the json.Unmarshaler interface,
+// keys will parse any value specified as a key to a string
+func (k *Key) UnmarshalJSON(data []byte) error {
+	if len(data) < 2 {
+		k.Clean()
+		return errors.New("too short to unmarshal key to json string")
+	}
+	*k = NewKey(string(data[1 : len(data)-1]))
```
whyrusleeping: Whatever escaping we do in the marshal step, we should do the opposite here

key.go
Stebalien: Any reason to not just call `json.Marshal(k.String())`?

key.go
Stebalien: Again, why not just use the `json` package? Doing our own parsing sounds like a bad idea.

Stebalien (commented):
I'm not entierly sure what this has to do with IPLD but I see no reason not to do this. However, I'm strongly against custom json parsing/formatting. We should just let the json package do the heavy lifting.

b5:
@Stebalien thanks for the vote in support of importing the json package directly, I completely agree that a custom implementation is less optimal. I'd initially written a custom implementation because go-datastore is something of an "interface package", with concrete implementations delegated to other / sub packages.

Personally, I like interface packages to be as lightweight as possible, importing `encoding/json` represents something of a beachhead for importing other encoding packages. Later contributors may say "well, we have json imported, why not import `encoding/gob` as well?", I don't want to be the person that gets that ball rolling 😉.

That being said, the response to that question could just be "because json is more popular", and if that's sufficient for y'all, I'll flip this PR to using std json encoding ASAP. Again, completely agree that `encoding/json` is the better way to go.

The other option worth consideration is just changing `type Key struct { key string }` to `type Key string` (I mentioned this in my initial comment). Encoders that (properly) use reflection treat types derived directly from a string as a string. The reason I haven't pushed harder for that option is a lack of familiarity with how this would affect other code.

Playing devil's advocate to this option, changing Key to be a string type would allow users to do this:
```go
datastore.Key("this might not be a valid key")
```
Which would allow users an easy way to get around the call to `key.Clean()` that's automatically invoked with `NewKey`, and skipped with `RawKey`, which in the long run could require more defensive coding from things that accept keys. I'm guessing this is a major reason for not going this route. I'd personally recommend conquering this with documentation that says all users _must_ always use NewKey, which is consistent with go idioms.

Apologies for the big dust-up on just getting an effing key into JSON, I promise it's because we need this for a number of really good reasons and I don't want to ruin an artfully composed package.

whyrusleeping:
> this with documentation that says all users must always use NewKey, which is consistent with go idioms.

The issue here was performance, switching to a mode where we can trust the keys and allow for `RawKey` to be called saved use huge amounts of cpu and memory pressure (just ask @kevina )

b5:
roger that, I've updated the PR to use standard json package methods. I've left the tests in, but would be happy to rip them out if they feel extraneous. At this point they're not testing much of value.

Seems the go-tip tests are failing, which is a little confusing.

whyrusleeping (approved):

Stebalien:
I've rebased off master. All tests should now pass (hopefully); I hate merging things with red Xs.

Stebalien (approved):

PR #71, opened 2017-11-05
Fix warnings and errors on go-tip

Stebalien:


Stebalien:
Note: There are still warnings about public functions returning private types. However, returning interfaces breaks some of our tests and I don't know if the types should actually be exported.

Stebalien:
Probably doesn't need a review...

PR #72, opened 2018-01-08
update go.uuid

Stebalien:
It now returns entropy errors so we now need to handle them. Lovely.

### DIFF
--- a/key.go
+++ b/key.go
@@ -254,7 +254,7 @@ func (k *Key) UnmarshalJSON(data []byte) error {
 //   RandomKey()
 //   NewKey("/f98719ea086343f7b71f32ea9d9d521d")
 func RandomKey() Key {
-	return NewKey(strings.Replace(uuid.NewV4().String(), "-", "", -1))
+	return NewKey(strings.Replace(uuid.Must(uuid.NewV4()).String(), "-", "", -1))
 }
 
 /*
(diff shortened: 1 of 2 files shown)


frrist (approved):

key.go
```diff
@@ -254,7 +254,7 @@ func (k *Key) UnmarshalJSON(data []byte) error {
 //   RandomKey()
 //   NewKey("/f98719ea086343f7b71f32ea9d9d521d")
 func RandomKey() Key {
-	return NewKey(strings.Replace(uuid.NewV4().String(), "-", "", -1))
+	return NewKey(strings.Replace(uuid.Must(uuid.NewV4()).String(), "-", "", -1))
```
Kubuxu, line 257: I think using `Must` here is not the best idea as it will lead to panic in case of low entropy. Instead we might want to retry and/or wait+retry.

Stebalien: Doesn't that only apply to e.g. embedded devices? We should probably do that anyways but, IIRC, the only difference is that the old constructor paniced internally.

Kubuxu: It would be also a problem on multi-tenant servers (high entropy usage) or on SSL terminating servers (with high tragic) without Haveged.

Stebalien:
> It would be also a problem on multi-tenant servers (high entropy usage) or on SSL terminating servers (with high tragic) without Haveged.

This should only be an issue on early boot on old Linux kernels (<3.17) and other Unixes. I'm not sure about windows but looks like it always works unless something is really wrong.

PR #74, opened 2018-01-18
Add Size() method to datastores

hsanjuan:
The each implementation of a datastore should decide on their size
is measured and how accurate the result is.

i.e. Some datastores may consider Size as the number of entries,
others as the disk space they use etc. Some implementations may
opt to improve the performance of the operation by caching or
by reducing the accuracy of the calculation.

Related: https://github.com/ipfs/go-ipfs/pull/4550

### DIFF
--- a/basic_ds.go
+++ b/basic_ds.go
@@ -168,9 +168,8 @@ func (d *LogDatastore) Delete(key Key) (err error) {
 
 // DiskUsage implements the PersistentDatastore interface.
 func (d *LogDatastore) DiskUsage() (uint64, error) {
-	du, err := DiskUsage(d.child)
-	log.Printf("%s: DiskUsage: %d\n", d.Name, du)
-	return du, err
+	log.Printf("%s: DiskUsage\n", d.Name)
+	return DiskUsage(d.child)
 }
 
 // Query implements Datastore.Query
--- a/basic_ds.go
+++ b/basic_ds.go
@@ -166,6 +166,13 @@ func (d *LogDatastore) Delete(key Key) (err error) {
 	return d.child.Delete(key)
 }
 
+// DiskUsage implements the PersistentDatastore interface.
+func (d *LogDatastore) DiskUsage() (uint64, error) {
+	du, err := DiskUsage(d.child)
+	log.Printf("%s: DiskUsage: %d\n", d.Name, du)
+	return du, err
+}
+
 // Query implements Datastore.Query
 func (d *LogDatastore) Query(q dsq.Query) (dsq.Results, error) {
 	log.Printf("%s: Query\n", d.Name)
--- a/callback/callback.go
+++ b/callback/callback.go
@@ -40,3 +40,9 @@ func (c *Datastore) Query(q dsq.Query) (dsq.Results, error) {
 	c.F()
 	return c.D.Query(q)
 }
+
+// DiskUsage implements the PersistentDatastore interface.
+func (c *Datastore) DiskUsage() (uint64, error) {
+	c.F()
+	return ds.DiskUsage(c.D)
+}
--- a/coalesce/coalesce.go
+++ b/coalesce/coalesce.go
@@ -138,3 +138,8 @@ func (d *datastore) Close() error {
 	}
 	return nil
 }
+
+// DiskUsage implements the PersistentDatastore interface.
+func (d *datastore) DiskUsage() (uint64, error) {
+	return ds.DiskUsage(d.child)
+}
--- a/datastore.go
+++ b/datastore.go
@@ -95,7 +95,7 @@ type CheckedDatastore interface {
 
 // CheckedDatastore is an interface that should be implemented by datastores
 // which want to provide a mechanism to check data integrity and/or
-// error correction
+// error correction.
 type ScrubbedDatastore interface {
 	Datastore
 
@@ -103,11 +103,31 @@ type ScrubbedDatastore interface {
 }
 
 // GCDatastore is an interface that should be implemented by datastores which
-// don't free disk space by just removing data from them
+// don't free disk space by just removing data from them.
 type GCDatastore interface {
 	Datastore
 
 	CollectGarbage() error
+}
+
+// PersistentDatastore is an interface that should be implemented by datastores
+// which can report disk usage.
+type PersistentDatastore interface {
+	Datastore
+
+	// DiskUsage returns the space used by a datastore, in bytes.
+	DiskUsage() (uint64, error)
+}
+
+// DiskUsage checks if a Datastore is a
+// PersistentDatastore and returns its DiskUsage(),
+// otherwise returns 0.
+func DiskUsage(d Datastore) (uint64, error) {
+	persDs, ok := d.(PersistentDatastore)
+	if !ok {
+		return 0, nil
+	}
+	return persDs.DiskUsage()
 }
 
 // Errors
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -1,5 +1,5 @@
 // Package fs is a simple Datastore implementation that stores keys
-// are directories and files, mirroring the key. That is, the key
+// as directories and files, mirroring the key. That is, the key
 // "/foo/bar" is stored as file "PATH/foo/bar/.dsobject".
 //
 // This means key some segments will not work. For example, the
@@ -20,6 +20,7 @@ package fs
 import (
 	"fmt"
 	"io/ioutil"
+	"log"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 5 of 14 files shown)


kevina:
I would rather the definition be better defined as either the approximate number of entries or the approximate disk space used.

Stebalien:
Yeah, we don't usually care about the size of the datastore. We care about the disk usage. I was thinking something like:

```go
type PersistentDatastore interface {
    DiskUsage() (int64, error)
}
```

Where datastores could *optionally* implement this. Implementing this interface *wouldn't* mean that the datastore necessarily persists its data, just that it understands the concept and can report any disk space it thinks it's using.

I got stuck with this interface trying to figure out a more generic "stats" interface but I think we can just go with something like the above and deal with other stats later.

hsanjuan:
ok, but if a persistent datastore is wrapped in any of the wrapper ones, then your "optional" method is gone, and you need to cast (which makes the interface less useful and not very helpful in the case of triages etc).

We could just force everyone to report `DiskUsage` and those that don't care about it should just say 0...

hsanjuan:
> we don't usually care about the size of the datastore

It would not hurt allowing to know this information though. In a datastore where all entries are roughly the same size (thinking ipfs chunks here) it can offer a pretty good ballpark of the total size with little effort...

Stebalien:
> ok, but if a persistent datastore is wrapped in any of the wrapper ones, then your "optional" method is gone, and you need to cast (which makes the interface less useful and not very helpful in the case of triages etc).

Datastore wrappers would have to handle this case. We're already using this pattern for, e.g., GC.

> It would not hurt allowing to know this information though. In a datastore where all entries are roughly the same size (thinking ipfs chunks here) it can offer a pretty good ballpark of the total size with little effort...

If we don't agree on what size means, it is an *entirely* useless metric. What if I have a mapped datastore that delegates to two datastores, one that treats size as the number of objects and another that treats size as bytes? What about caches?

hsanjuan:
> If we don't agree on what size means, it is an entirely useless metric. 

What I mean is, exposing the number of elements under a different method too (`Size` and `DiskUsage` for example), since we're at it.

> Datastore wrappers would have to handle this case. We're already using this pattern for, e.g., GC.

ok I will proceed that way then

hsanjuan:
How is this looking?

datastore.go
```diff
@@ -110,6 +110,14 @@ type GCDatastore interface {
 	CollectGarbage() error
 }
 
+// PersistentDatastore is an interface that should be implemented by datastores
+// which can report disk usage.
+type PersistentDatastore interface {
+	Datastore
+
+	DiskUsage() (uint64, error)
```
magik6k: Should have a doc explicitly saying what the unit here is (i.e. bytes).

magik6k (commented):
Code LGTM, one comment

datastore.go
kevina: I would think that would be obvious, but it can't hurt.

basic_ds.go
```diff
@@ -166,6 +166,16 @@ func (d *LogDatastore) Delete(key Key) (err error) {
...
 
+func (d *LogDatastore) DiskUsage() (uint64, error) {
+	var du uint64 = 0
+	var err error
+	if pd, ok := d.child.(PersistentDatastore); ok {
+		du, err = pd.DiskUsage()
+	}
+	log.Printf("%s: DiskUsage: %d\n", d.Name, du)
+	return du, err
+}
```
kevina: This function should return an error if the child is not a PersistentDatastore.

I would also restructure it to avoid having to declare du and err up front.

Something like
```go
pd, ok := d.child.(PersistentDatastore)
if !ok {
  return 0, errors.New("Unimplemented")
}
du, err := pd.DiskUsage()
if err != nil {
  return 0, err
}
log.Printf("%s: DiskUsage: %d\n", d.Name, du)
return du, nil
```
But use a better error message.

kevina (changes requested):
If the `DiskUsage` method is not available an error should be returned.  See inlined comments.

datastore.go
hsanjuan: yeah you're totally right, my bad

basic_ds.go
hsanjuan: I see why a batch operation for example, should return an error if the underlying datastore cannot perform it at all. But DiskUsage? This implies that mixing Persistent and non PersistentDatastores in a wrapper and asking for DiskUsage should probably result in error too (as returning error here would mean doing it everywhere for behavior consistency).

If the datastore is not persistent, it seems natural to say (from a wrapper point of view) that the DiskUsage is 0, while keeping the error result to actually signal for real errors when obtaining real disk usage. In the current approach wrappers never produce errors on DiskUsage(), they just bubble them.

callback/callback.go
```diff
@@ -40,3 +40,14 @@ func (c *Datastore) Query(q dsq.Query) (dsq.Results, error) {
...
 }
+
+// DiskUsage returns the size if the child datastore, or 0 if it's not
+// a PersistentDatastore.
+func (c *Datastore) DiskUsage() (uint64, error) {
+	c.F()
+	if pd, ok := c.D.(ds.PersistentDatastore); ok {
+		return pd.DiskUsage()
+	}
+
```
kevina: See comments from previous review.

I would also refactor the logic info a helper function.

(Now GitHub is annoying me, This is the third time I wrote this comment.!)

coalesce/coalesce.go
```diff
@@ -138,3 +138,12 @@ func (d *datastore) Close() error {
...
 	return nil
 }
+
+// DiskUsage returns the size if the child datastore, or 0 if it's not
+// a PersistentDatastore.
+func (d *datastore) DiskUsage() (uint64, error) {
+	if pd, ok := d.child.(ds.PersistentDatastore); ok {
+		return pd.DiskUsage()
+	}
+	return 0, nil
```
kevina: Ditto.

Etc.

callback/callback.go
hsanjuan: Would you create a separate util module just for this helper?

basic_ds.go
kevina: If the underlying datastore uses disk storage but doesn't implement this method, then a result of 0 without an error would be misleading.  

Maybe we should just require this method be implemented, or perhaps a more generic Stat() method.

@Stebalien thoughts?

callback/callback.go
kevina: I would just put in in the top-level package and give it would look something like
```
func DiskUsage(d *Datastore) (uint64, error) {
	pd, ok := d.(PersistentDatastore)
	if !ok {
		return 0, errors.New("Unimplemented")
	}
	du, err := pd.DiskUsage()
	if err != nil {
		return 0, err
	}
	return du, nil
}
```
And then in the wrapper methods just use:
```
return DiskUsage(c.D)
```

hsanjuan: ok

basic_ds.go
Stebalien: If a datastore uses the disk, it should report it. If it doesn't, that's a bug in the datastore. However, I don't want to force every datastore to implement this method (it's not really necessary).

I'd like to go with a more generic stat but I couldn't come up with a concrete, nice solution that'll please everyone so, unless you can think of a way to make that work (that's both efficient and idiomatic) we might as well go with this.

callback/callback.go
Stebalien: I agree with having a helper function to avoid casting everywhere. I disagree with not implementing this interface an error.

basic_ds.go
kevina: Okay, I am unconformable with the idea of unimplementable this method means it doesn't use any disk space, but I do agree that will be easier.

I take another look but I think this LGTM then.

callback/callback.go
kevina: (Note this will likely be overkill, if we decide not to return an error.)

hsanjuan:
Thanks for your feedback and patience everyone! Ready for another round...

Stebalien (approved):

Stebalien:
@kevina, @magik6k LGTY?

basic_ds.go
```diff
@@ -166,6 +166,13 @@ func (d *LogDatastore) Delete(key Key) (err error) {
 	return d.child.Delete(key)
 }
 
+// DiskUsage implements the PersistentDatastore interface.
+func (d *LogDatastore) DiskUsage() (uint64, error) {
+	du, err := DiskUsage(d.child)
+	log.Printf("%s: DiskUsage: %d\n", d.Name, du)
```
kevina: This should probably not report anything if an error was returned.

kevina (changes requested):
One minor change.  Otherwise LGTM.

We need to make sure that this new method is implemented on all Datastores before this method is used.  (I would still have preferred an error be returned if it wasn't implemented.)

basic_ds.go
hsanjuan: The rest of methods there log the calls regardless of results. I can log the call and not the result though (just did that)

kevina (approved):
LGTM.

magik6k (approved):

hsanjuan:
Will someone be so kind to merge?

kevina:
@hsanjuan I believe that would be @whyrusleeping or @Stebalien job.

Stebalien:
@hsanjuan are you going to put together the implementations?

hsanjuan:
@Stebalien yeah that's my intention

Issue #75, opened 2018-01-23
Figure out what's going on with the locking in syncmount


Stebalien:
The datastores in syncmount use their locks inconsistently.

1. Mount doesn't appear to actually need a lock. Is this correct?
2. mountBatch does modify it's mounts map but fails to lock around some methods (e.g., commit, check, scrub, etc).

Issue #76, opened 2018-01-23
Delete old, unused datastores.


Stebalien:
Having to maintain old, unused, untested code is a pain (they won't work anyways because most of them haven't really been tested).

magik6k:
List of what is actually used and how (chocked by searching for usages on latest gx version used by go-ipfs):
* `callback`
  * Used in `coalesce` datastore tests (which is not used)
  * Used in `tiered` datastore tests (which is not used)
* `coalesce`
  * doesn't appear to be used by anything
* `elastigo`
  * ...
  * not used
* `fs`
  * not used
  * we have flatfs
  * it's a tiny, simple implementation of persistent datastore which can be used by anyone who wants to use this package
    * it may be buggy
    * we have flatfs
* `keytransform`
  * used by `namespace`
* `mount`
  * not used
* `namespace`
  * Used by `mount`, `syncmount`, in go-ipfs for `blackstore`, `filestore`
* `panic`
  * not used
  * ...
  * I mean, it may be useful to debug some weird issue, though I'd much rather just use conditional breakpoints.
* `sync`
  * used all over the place in go-ipfs
* `syncmount`
  * used in go-ipfs
* `tiered`
  * not used
  * /might/ be useful, though it doesn't implement batching so I'd say we should drop it, and if anyone wants it back, VCS will still have it.
* `timecache`
  * Same as above

Datastores in basic_ds.go:
* `MapDatastore`:
  * used in many tests everywhere
* `NullDatastore`:
  * Used in some tests
* `LogDatastore`:
  * Usable in go-ipfs (though mdns needs to be disabled due to a bug)

So I'd say we should remove `callback`, `coalesce`, `elastigo`, `fs`, `mount`, `tiered` and `timecache`, @Stebalien LGTY?

magik6k:
And probably just rename `syncmount` to `mount`

Stebalien:
LGTM. @whyrusleeping? You should probably sign off on this since we're deleting code.

whyrusleeping:
@magik6k @Stebalien lets make a tag on the commit the code exists in before we delete it, and note it in an issue somewhere. After that, SGTM

kevina:
Some of those might be useful as examples, such as `fs` and may be worth keeping around.  Maybe we can move those to the examples directory.

Stebalien:
[...]

PR #77, opened 2018-01-24
Fix lock usage in syncmount

magik6k:
Fixes https://github.com/ipfs/go-datastore/issues/75

To answer questions:

> Mount doesn't appear to actually need a lock. Is this correct?

`datastore.mounts` are immutable, so yes, this appears to be correct.

> mountBatch does modify it's mounts map but fails to lock around some methods (e.g., commit, 
check, scrub, etc).

Commit should only be called after the batch has been finished, though adding the lock there shouldn't hurt.

Additionally note that check/scrub/gc are methods on datastore, not batch

### DIFF
--- a/syncmount/mount.go
+++ b/syncmount/mount.go
@@ -35,14 +35,11 @@ func New(mounts []Mount) *Datastore {
 
 type Datastore struct {
 	mounts []Mount
-	lk     sync.Mutex
 }
 
 var _ ds.Datastore = (*Datastore)(nil)
 
 func (d *Datastore) lookup(key ds.Key) (ds.Datastore, ds.Key, ds.Key) {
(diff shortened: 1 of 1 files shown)


Stebalien:
> Commit should only be called after the batch has been finished, though adding the lock there shouldn't hurt.

Can we not commit multiple times? (if so, we might want to poison the datastore in that case.

Also, do we actually *need* the mount datastore? We already have one in `mount/mount.go`.

Stebalien (approved):
LGTM

magik6k:
> Also, do we actually need the mount datastore? We already have one in `mount/mount.go`.

See https://github.com/ipfs/go-datastore/issues/76#issuecomment-359980154 (TL;DR: No)

PR #78, opened 2018-01-26
Delete old, unused datastores

### LINKED ISSUE #76: Delete old, unused datastores.

magik6k:
Closes #76 

* Removed datastores: `callback`, `coalesce`, `elastigo`, `mount`, `panic`, `tiered`, `timecache`
* Moved `fs` to examples/
* Q: should I rename `syncmount` to `mount`?

I have pushed a tag with current master to have easier access to the removed datastores in case anyone wants to make them more complete.


Stebalien:
LGTM (modulo merge conflicts). Just to tripple check, nothing uses these, right? Are there any changes we need to make higher up the stack?

magik6k:
Rebased

> Just to tripple check, nothing uses these, right? Are there any changes we need to make higher up the stack?

IF we don't decide to rename `syncmount` to `mount` then there is nothing to change (if we do then it's likely the matter of updating [this](https://github.com/ipfs/go-ipfs/blob/master/repo/fsrepo/datastores.go#L17) import).

Stebalien:
Got it. Let's rename it now. That's only one change and a simple one at that.

magik6k:
Done

Stebalien:
Awesome! Thanks!.

PR #81, opened 2018-02-09
Import ds-help methods from go-ipfs/thirdparty

hsanjuan:
There is a TODO: put this code into the go-datastore itself.

I will PR to ipfs after this is merged and a new gx published.

### DIFF
--- a/ds_help.go
+++ b/ds_help.go
@@ -1,0 +1,29 @@
+package datastore
+
+import (
+	cid "github.com/ipfs/go-cid"
+	"github.com/whyrusleeping/base32"
+)
+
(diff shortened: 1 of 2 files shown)


magik6k (commented):
go-datastore is not specific to IPFS, I'm not sure if we want these methods here.

hsanjuan:
I can extract it to a separate repo. @whyrusleeping, you left that TODO. What do you think?

whyrusleeping:
@hsanjuan Yeah, the weirdness is putting cids into go-datastore. I don't think they belong there.

hsanjuan:
I will extract this to its own repo no problem :)

Issue #94, opened 2018-08-13
Make the datastores take/return []byte slices


Stebalien:
This has been bugging me for a while so I figured I'd just go ahead and do it.

kevina:
@Stebalien there might have been a reason for the {}interface return value.  I would get the all clear from @whyrusleeping (and maybe jbenet, but don't want to bother him unless necessary) before proceeding as he would know the historical context better then either of us.  My apologizes if you already did.

Stebalien:
Thanks for bringing that up but I have talked with juan about this. In theory, the "datastore" was supposed to support arbitrary objects. However, that doesn't work well in practice and we never actually used that feature. I'm going to get @whyrusleeping's :+1: when CI is green.

kevina:
Okay as long as Juan and @whyrusleeping are okay with it, it sounds good to me.

Stebalien:
Last time I checked, Juan was more resigned to it than ok but close enough.

whyrusleeping:
Yolo

Stebalien:
All done!

PR #95, opened 2018-08-13
go datastore {}interface -> []byte refactor

Stebalien:
See: https://github.com/ipfs/go-datastore/issues/94

### DIFF
--- a/query/query.go
+++ b/query/query.go
@@ -64,16 +64,10 @@ type Query struct {
 	KeysOnly bool     // return only keys.
 }
 
-// NotFetched is a special type that signals whether or not the value
-// of an Entry has been fetched or not. This is needed because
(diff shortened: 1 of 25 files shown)


query/query.go
```diff
@@ -64,16 +64,10 @@ type Query struct {
 	KeysOnly bool     // return only keys.
 }
 
-// NotFetched is a special type that signals whether or not the value
-// of an Entry has been fetched or not. This is needed because
-// datastore implementations get to decide whether Query returns values
-// or only keys. nil is not a good signal, as real values may be nil.
-const NotFetched int = iota
```
Stebalien, line 71: We weren't actually using this anywhere.

bigs:
after the fact, but love this :)

PR #96, opened 2018-09-12
Add support for querying entry expiration

raulk:
Adds the following:

* `GetExpiration` function on `TTLDatastore`.
* `ReturnExpirations` field on query options.
* `Expiration` on query result object.

The only implementation of the affected interface is go-ds-badger, and the PR to implement this functionality is here: https://github.com/ipfs/go-ds-badger/pull/32

### DIFF
--- a/query/query.go
+++ b/query/query.go
@@ -1,6 +1,8 @@
 package query
 
 import (
+	"time"
+
 	goprocess "github.com/jbenet/goprocess"
 )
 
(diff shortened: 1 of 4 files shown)


Stebalien (approved):
LGTM but we should get either @bigs or @magik6k to sign off on this as well.

bigs (approved):

query/query.go
```diff
@@ -56,18 +58,20 @@ cost of the layer of abstraction.
...
-	Limit    int      // maximum number of results
-	Offset   int      // skip given number of results
-	KeysOnly bool     // return only keys.
+	Prefix            string   // namespaces the query to results whose keys have Prefix
+	Filters           []Filter // filter results. apply sequentially
+	Orders            []Order  // order results. apply sequentially
+	Limit             int      // maximum number of results
+	Offset            int      // skip given number of results
+	KeysOnly          bool     // return only keys.
+	ReturnExpirations bool     // return expirations (see TTLDatastore)
```
magik6k, line 67: Maybe `GetExpirations`?

raulk: @magik6k I considered `GetExpirations`, but discarded it because it implies that the impl may perform an extra lookup to populate the field, when in reality the spirit is to allow returning this value if its available in the iterator.

raulk:
Thanks all. Holding off on merge until https://github.com/ipfs/go-ds-badger/pull/32 is approved, to avoid breaking interfaces.

PR #99, opened 2018-10-04
add a GetSize method

Stebalien:
We can also add this as an extension but it's simple enough I'd like to add it directly to the Datastore.

Use-case: The blockstore now has a GetSize and bitswap uses it for packing multiple blocks into a single message. Unfortunately, this means that bitswap is now calling blockstore.GetSize which calls datastore.Get which tends to be more expensive than datastore.Has.

Sub-PRs:

* Badger: https://github.com/ipfs/go-ds-badger/pull/38
* Leveldb: https://github.com/ipfs/go-ds-leveldb/pull/18
* Flatfs: https://github.com/ipfs/go-ds-flatfs/pull/48
* Measure: https://github.com/ipfs/go-ds-measure/pull/17

### DIFF
--- a/.gx/lastpubver
+++ b/.gx/lastpubver
@@ -1,1 +1,1 @@
-3.3.0: QmbQshXLNcCPRUGZv4sBGxnZNAHREA6MKeomkwihNXPZWP
+3.4.0: QmaRb5yNXKonhbkpNxNawoydk4N6es6b4fPj19sjEKsh5D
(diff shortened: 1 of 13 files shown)


kevina (approved):
LGTM

magik6k (approved):

bigs (approved):

hsanjuan (approved):

schomatis (approved):

bigs (approved):

PR #108, opened 2018-12-26
delayed: implement io.Closer and export datastore type.

Stebalien:


delayed/delayed.go
```diff
@@ -67,3 +70,11 @@ func (dds *Delayed) DiskUsage() (uint64, error) {
 	dds.delay.Wait()
 	return ds.DiskUsage(dds.ds)
 }
+
+// Close closes the inner datastore (if it implements the io.Closer interface).
+func (dds *Delayed) Close() error {
+	if closer, ok := dds.ds.(io.Closer); ok {
```
schomatis, line 76: So, if I understand correctly, this will be forward compatible with a `Close` method that `D` is not implementing at the moment, right?

schomatis (approved):
LGTM

delayed/delayed.go
Stebalien: This is for the underlying datastore, not the delay. It allows us to wrap e.g. badger without hiding the `Close` method.

PR #110, opened 2019-01-24
query: make datastore ordering act like a user would expect

Stebalien:
This applies orders hierarchically. Unfortunately, the interfaces had to be changed significantly to get this to work.

For example `[OrderByValue{}, SortByKey{}]` *used* to re-sort by key, throwing away the value sort. Now, this acts like a normal database and `SortByKey` will only sort within equivalent values.

**[[BREAKING CHANGE]]**

PRs for datastores:

* go-ds-badger - https://github.com/ipfs/go-ds-badger/pull/44
* go-ds-flatfs - doesn't even support this.
* go-ds-leveldb - https://github.com/ipfs/go-ds-leveldb/pull/23
* sql-datastore - https://github.com/whyrusleeping/sql-datastore/pull/6

(and probably more)

This is all because we had to change `NaiveOrder`. Unfortunately, the old interface simply can't support this kind of sorting.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,6 @@
   "license": "MIT",
   "name": "go-datastore",
   "releaseCmd": "git commit -a -m \"gx publish $VERSION\"",
-  "version": "3.4.1"
+  "version": "3.5.0"
 }
 
--- a/query/order.go
+++ b/query/order.go
@@ -1,66 +1,48 @@
 package query
 
 import (
-	"sort"
+	"bytes"
+	"strings"
 )
 
 // Order is an object used to order objects
 type Order interface {
-
-	// Sort sorts the Entry slice according to
-	// the Order criteria.
-	Sort([]Entry)
+	Compare(a, b Entry) int
 }
 
-// OrderByValue is used to signal to datastores they
-// should apply internal orderings. unfortunately, there
-// is no way to apply order comparisons to interface{} types
-// in Go, so if the datastore doesnt have a special way to
-// handle these comparisons, you must provide an Order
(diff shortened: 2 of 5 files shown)


query/order.go
```diff
@@ -1,66 +1,48 @@
...
 // OrderByValueDescending is used to signal to datastores they
-// should apply internal orderings. unfortunately, there
-// is no way to apply order comparisons to interface{} types
-// in Go, so if the datastore doesnt have a special way to
-// handle these comparisons, you are SOL.
-type OrderByValueDescending struct {
-	TypedOrder Order
-}
+// should apply internal orderings.
+type OrderByValueDescending struct{}
```
raulk, line 30: Not related to the changes introduced herein, but since were making breaking changes, I think descending could just be an option of `OrderByKey` and `OrderByValue`.

package.json
```diff
@@ -37,6 +37,6 @@
   "license": "MIT",
   "name": "go-datastore",
   "releaseCmd": "git commit -a -m \"gx publish $VERSION\"",
-  "version": "3.4.1"
+  "version": "3.5.0"
```
raulk, line 40: Should we bump up the major version? Although admittedly nobody is using the ordering.

query/order.go
Stebalien: Note: We're not breaking end users, just datastore implementations. Otherwise, I agree.

package.json
Stebalien: We should... but as we're only breaking datastore implementations, I think we can get away with this.

raulk (approved):

PR #113, opened 2019-01-29
remove io.Closer from the transaction interface

### LINKED ISSUE #24: "go get -u github.com/ipfs/go-ds-leveldb" fails with error "cannot use d (type *accessor) as type datastore.Datastore in argument to datastore.GetBackedSize"
```bash
$ go get -u github.com/ipfs/go-ds-leveldb
# github.com/ipfs/go-ds-leveldb
../../../../pkg/mod/github.com/ipfs/go-ds-leveldb@v1.2.1/datastore.go:99:25: cannot use d (type *accessor) as type datastore.Datastore in argument to datastore.GetBackedSize:
        *accessor does not implement datastore.Datastore (missing Close method)
../../../../pkg/mod/github.com/ipfs/go-ds-leveldb@v1.2.1/datastore.go:306:32: cannot use transaction literal (type *transaction) as type datastore.Txn in return argument:
        *transaction does not implement datastore.Txn (missing Close method)
```

```bash
$ go version
go version go1.11.5 darwin/amd64
```

Stebalien:
Transactions shouldn't have to implement Close.

fixes https://github.com/ipfs/go-ds-leveldb/issues/24

### DIFF
--- a/datastore.go
+++ b/datastore.go
@@ -171,7 +171,8 @@ type TTLDatastore interface {
 // Commit has been made. Likewise, transactions can be aborted by calling
 // Discard before a successful Commit has been made.
 type Txn interface {
-	Datastore
+	Read
+	Write
 
 	// Commit finalizes a transaction, attempting to commit it to the Datastore.
 	// May return an error if the transaction has gone stale. The presence of an


Stebalien:
Merging as this is an emergency fix to unbreak stuff.

PR #118, opened 2019-02-06
Splinter TTLDatastore interface into TTL + Datastore

raulk:
After introducing the Close method in Datastore, we no longer had an interface to encapsulate only TTL methods, which broke go-ds-badger tests.


raulk:
Added a few extra reviewers to take some stuff off @Stebalien's plate. This is pretty uncontroversial.

Stebalien (approved):
SGTM

bigs:
after-the-fact approval! looks nice in keeping with all other changes happening rn

Issue #116, opened 2019-02-04
No way to query past Query.Prefix?


drew-512:
Hi gents!

Currently, `Query.Prefix` is used to do the initial seek when a `Query` begins, and iteration only continues while that prefix holds.  There needs to be a flag that basically says, hey, don't use `Prefix` _once iteration has begun_.  Otherwise, there's no way to effectively do a seek and do an arbitrary amount of scanning.  This seems like a critical feature to leave behind, yikes.

Hopefully, I'm missing something!?

For example, in `go-ds-badger`:  (line 270+)

```golang
it := txn.NewIterator(opt)
it.Seek(prefix)
...
qrb.Process.Go(func(worker goprocess.Process) {
    ...
    for sent := 0; it.ValidForPrefix(prefix); sent++ {
        ...
    }
}
...
```

could be enhanced to:

```golang
it := txn.NewIterator(opt)
it.Seek(prefix)
if q.PrefixOnlySeeks {
    prefix = nil
}
...
```

Adding this would be easy, but all the impls would need to be updated of course.   So maybe it's done as a selectively available feature?  

This is a dealbreaker for us, but we love Datastore and want to use and support it!   If some at PL agrees to accept our pull requests in a timely fashion, we'll help do the work. I'm sad we're now in this position, and would take less than a few mins for dbs like leveldb, badger, and bolt.  What do y'all think?

We kindly need resolution on this relatively soon one way or the other so that we know we need to pass on Datastore.

Thanks,
Drew

magik6k:
Some datastores may not be able to support this (like flatfs, but it doesn't even support ordered iterating, so it's probably fine).

:+1: from me, just make sure that it's optional and not enabled by default (default = zero value)

cc @Stebalien @bigs

drew-512:
Ok, cool, thanks for the quick response.

How about I whip up some changes, post them here for initial review/discussion and if they look good, we can take it from there.

drew-512:
Ok, so so here's my proposed change to `Query`. 

`query.go` ~ line 60
```golang
type Query struct {
	Prefix            string   // namespaces the query to results whose keys have Prefix
	IterationPrefix   string   // if set, this overrides Prefix during query iteration (n/a for some impls)
```

[...]

PR #119, opened 2019-02-06
added Query.SeekPrefix

drew-512:


Stebalien:
There should probably be a naive implementation in query_impl.go

drew-512:
Something like:

```golang
type FilterKeySeekPrefix struct {
    SeekPrefix string
}

func (f FilterKeySeekPrefix) Filter(e Entry) bool {
	return e.Key >= f.SeekPrefix
}
```

and

```golang
func NaiveQueryApply(q Query, qr Results) Results {
	if q.Prefix != "" {
		qr = NaiveFilter(qr, FilterKeyPrefix{q.Prefix})
	}
	if q.SeekPrefix != "" {
		qr = NaiveFilter(qr, FilterKeySeekPrefix{q.SeekPrefix})
	}
	...
```

?

Stebalien:
That will just act like `Prefix`. This one will have to *skip* everything until `SeekPrefix`. It also needs to be applied after any ordering.

drew-512:
> That will just act like `Prefix`. This one will have to _skip_ everything until `SeekPrefix`. It also needs to be applied after any ordering.

Let's say we have entries whose keys are zero-padded time codes for speedy scanning by time:
`/txn/000009-DEADBEEF`
`/txn/000015-FEEDFACE`
`/txn/000055-BEEFBEEF`
`/txn/000055-B01DFACE`
`/txn/000103-D0D0CACA`

.SeekPrefix could be something like `/txn/000050` and .Prefix would be `/txn/`, so the thing I'm seeing is that `NaiveQueryApply` should look at `OrderByKeyDescending` (as you described in the other thread) in order to know which way to compare the key with `.SeekPrefix`:

```golang
type FilterKeySeekPrefix struct {
    Ascending bool
    SeekPrefix string
}

func (f FilterKeySeekPrefix) Filter(e Entry) bool {
	if f.Ascending {
		return e.Key >= f.SeekPrefix
	} else {
		return e.Key <= f.SeekPrefix
        }
}
```

Is the only place `NaiveQueryApply` is used is in example code?   Maybe it was helpful when demoing `Datastore` way back, but is it likely to ever see production?  I would have a sit down w/ someone under me who ever used it, but that's just me.   Not that these SeekPrefix changes should drive code to be dropped from ds, but is this candidate code to be dropped?   Disclaimer: I'm super OCD about non-production/non-scalable staying around, so feel free to ignore this -- I always like to ask this question. ¯\\_(ツ)_/¯

Stebalien:
You're right, assuming ordering by key, that should work. Actually, I'm starting to wonder how this should interact with sorting by value. Let's discuss here: https://github.com/ipfs/go-datastore/issues/116#issuecomment-461232579

> Is the only place NaiveQueryApply is used is in example code?

We use it in the `MapDatastore` (in-memory datastore) and many datastores use the component functions (e.g., `NaiveFilter`).

drew-512:
Ok cool, thanks for indulging me on that ques and glad you're catching this stuff.

Like that we're gettin through this together and adding a nice feature to ds.

sanderpick:
This looks pretty stale... does that mean another avenue was chosen? Or key seeking is still not possible?

PR #120, opened 2019-03-15
remove ThreadSafeDatastore

Stebalien:
It's a lie! We:

1. Assume that our datastores are thread-safe all over the place, not bothering to check for this interface.
2. Implement this interface for, e.g., the mount datastore that _may not_ be thread-safe (depending on the sub-datastores).

Basically, there's no sane way to to do something like this in go. What we _want_ is:

```rust
pub trait ThreadSafe {}

struct MyWrapper<D: Datastore> { ... }

impl<D: Datastore> ThreadSafe for MyWrapper<D> where D: ThreadSafe {}
```

Actually, we don't even need this because rust has already done all the hard
work with the `Sync` trait.

....

But we're using go which barely has types.

---

For completeness, it's actually possible to do this in go:

```go
type threadSafeMixin struct{}
func (threadSafeMixin) ThreadSafe() {}

func NewWrapper(d Datastore) Datastore {
  if _, ok := d.(ThreadSafe) {
    return &struct{myWrapper, threadSafeMixin}{myWrapper{d}, threadSafeMixin{}}
  }
  return &myWrapper{d}
}
```

Let's not.

---

Merge first:

* [x] https://github.com/ipfs/go-ds-flatfs/pull/53
* [x] https://github.com/ipfs/go-ds-redis/pull/4
* [x] https://github.com/ipfs/go-ds-badger/pull/55
* [x] https://github.com/ipfs/go-ds-leveldb/pull/27

### DIFF
--- a/basic_ds.go
+++ b/basic_ds.go
@@ -13,7 +13,9 @@ type MapDatastore struct {
 	values map[Key][]byte
 }
 
-// NewMapDatastore constructs a MapDatastore
+// NewMapDatastore constructs a MapDatastore. It is _not_ thread-safe by
+// default, wrap using sync.MutexWrap if you need thread safety (the answer here
+// is usually yes).
 func NewMapDatastore() (d *MapDatastore) {
 	return &MapDatastore{
 		values: make(map[Key][]byte),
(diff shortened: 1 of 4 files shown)


magik6k (approved):

hsanjuan:
Heads up that this was a breaking API change and was released under a patch version bump (consider doing a minor release bump in that case, as most of us don't expect builds to break after `go get -u=patch`.

Stebalien:
You're right, I should have done that. I'm too used to versions meaning nothing at this point...

PR #121, opened 2019-03-18
Allow ordering by key on mount query

michaelavila:
This work was done to support https://github.com/ipfs/go-ipfs/pull/6068/files#r266291056.

## Overview

Mount datastore used to not support key ordering in queries, now it does.

### DIFF
--- a/mount/mount.go
+++ b/mount/mount.go
@@ -49,6 +49,12 @@ func (d *Datastore) lookup(key ds.Key) (ds.Datastore, ds.Key, ds.Key) {
 	return nil, ds.NewKey("/"), key
 }
 
+type lookupResult struct {
+	ds    ds.Datastore
+	mount ds.Key
+	rest  ds.Key
+}
+
 // lookupAll returns all mounts that might contain keys that are descendant of <key>
 //
 // Matching: /ao/e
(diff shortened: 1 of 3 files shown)


Stebalien (commented):
Looking at this change has made me realize that I made a mistake when I said this was going to be really easy. We can have the following mounts:

* /foo
* /foo/bar

In that case, we need to process some of `/foo` until we hit `/foo/bar`, then we need to process `/foo/bar`, then we need to go back to `/foo`.

There are a bunch of ways to handle this but, IMO, the simplest is to:

1. Query in parallel.
2. Put the next item in each query result  (along with a pointer to the query result) into a priority queue.
3. Pull the first item off the priority queue, replacing it with the next item from the query.

This shouldn't be difficult, just _slightly_ more complicated.

michaelavila:
@Stebalien that makes a lot of sense, we hadn't considered such mounts. I'll work on a change. Thanks.

michaelavila:
I've added a simplistic and poorly factored solution that addresses the problem, but probably won't be merged as is. A more reasonable solution, I think, is to (1) make the queries all in parallel, and (2) do a k-way merge on the results (which is what I think @Stebalien suggested in IRC).

I'm looking into the better solution now, but I have a proposal on the PR for the original work (https://github.com/ipfs/go-ipfs/pull/6068/files#r267463222) to consider pursuing this as its own task in order to get that other work merged.

michaelavila:
Superseded by https://github.com/ipfs/go-datastore/pull/124

PR #124, opened 2019-03-22
Fix – sorted limited offset mount queries

michaelavila:
Based on: https://github.com/ipfs/go-datastore/pull/123
Supersedes: https://github.com/ipfs/go-datastore/pull/121

Use NaiveLimit and NaiveOffset to add limit/offset features to queries on mount.

### DIFF
--- a/mount/mount.go
+++ b/mount/mount.go
@@ -222,10 +222,10 @@ func (d *Datastore) Delete(key ds.Key) error {
 
 func (d *Datastore) Query(master query.Query) (query.Results, error) {
 	childQuery := query.Query{
-		Prefix: master.Prefix,
-		Limit:  master.Limit,
-		Orders: master.Orders,
-		KeysOnly: master.KeysOnly,
+		Prefix:            master.Prefix,
+		Limit:             master.Limit,
+		Orders:            master.Orders,
+		KeysOnly:          master.KeysOnly,
 		ReturnExpirations: master.ReturnExpirations,
 	}
 
--- a/mount/mount.go
+++ b/mount/mount.go
@@ -3,6 +3,7 @@
 package mount
 
 import (
+	"container/heap"
 	"errors"
 	"fmt"
 	"sort"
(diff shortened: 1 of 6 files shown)


mount/mount.go
```diff
@@ -122,72 +227,40 @@ func (d *Datastore) Delete(key ds.Key) error {
...
+		}
+		queries.addResults(mount, results)
+	}
+
+	qr := query.ResultsFromIterator(q, query.Iterator{
+		Next:  queries.next,
+		Close: queries.close,
+	})
+
+	return query.NaiveLimit(query.NaiveOffset(qr, offset), q.Limit), nil
```
michaelavila: This works, but I think we want limit/offset implemented in a manner similar to filters. Is that true?

michaelavila:
I'm working on more tests now.

mount/mount.go
michaelavila: Maybe these things should only be applied if the `Limit` or `Offset` are non-zero?

eingenito: @michaelavila and I were just talking and we thought that maybe we should only wrap the results in NaiveLimit and NaiveOffset if those clauses exist in the Query to preserve the performance of existing code that obviously can't have ever used either of them.

michaelavila: The code is actually updated to only apply limit/filter when limit or filter are greater than zero.

mount/mount.go
```diff
@@ -49,6 +50,112 @@ func (d *Datastore) lookup(key ds.Key) (ds.Datastore, ds.Key, ds.Key) {
...
+	next := head.next
+
+	for head.advance() {
+		if head.next.Error == nil {
+			for _, f := range h.query.Filters {
+				if !f.Filter(head.next.Entry) {
+					continue
+				}
+			}
+			// can limit and offset be implemented here?
```
Stebalien: Limit, yes. For offset, we just have to use the naive solution.

However, using the naive limit should be fine _anyways_.

query/query_impl.go
```diff
@@ -137,6 +124,7 @@ func NaiveQueryApply(q Query, qr Results) Results {
 		qr = NaiveOffset(qr, q.Offset)
 	}
 	if q.Limit != 0 {
+		// TODO: Offset?
```
Stebalien: Yeah, that's broken... You can tell how often we use these...

mount/mount.go
```diff
@@ -49,6 +50,112 @@ func (d *Datastore) lookup(key ds.Key) (ds.Datastore, ds.Key, ds.Key) {
...
+	head := h.heads[0]
+	next := head.next
+
+	for head.advance() {
+		if head.next.Error == nil {
+			for _, f := range h.query.Filters {
+				if !f.Filter(head.next.Entry) {
+					continue
+				}
+			}
```
Stebalien: Honestly, we could probably switch to the naive filter as well. I'm not sure why I did it this way (well, it'll be slightly faster but not by much).

Stebalien (commented):
(sorry for leaving you with the annoying part)

mount/mount.go
eingenito: @Stebalien Actually we're finding that we might _have_ to use the NaiveFilter implementation because if the underlying datastores support filter they're going to be filtering only on the subkey (if they filter on key) meaning they'll exclude their mount prefix. So we probably have to treat filters in the same way we do offsets by stashing them and then removing them from the Query that is passed to each datastore, and then applying them at then end when the keys are full keys. So we're going to do as you suggest here and use NaiveFilter. It reads a little clearer too.

michaelavila: There's a commit that fixes filtering by using `NaiveFilter`: https://github.com/ipfs/go-datastore/pull/124/commits/5b4de5b291b89f2c73d5e391f931acc556f0b2e6

Stebalien: Ah, good catch. We definitely need to remove the filters from the query we're forwarding. Actually, we should probably just create a new query.

query/query_impl.go
michaelavila: Fixed and added tests.

query/query.go
```diff
@@ -3,7 +3,7 @@ package query
 import (
 	"time"
 
-	goprocess "github.com/jbenet/goprocess"
+	"github.com/jbenet/goprocess"
```
michaelavila: I'll fix this ...

Stebalien (approved):
LGTM!

PR #126, opened 2019-04-09
filter: values are now always bytes

Stebalien:


Stebalien:
We used to support values with arbitrary types but this added more complexity than it helped (we never actually _used_ arbitrary types, we just had a bunch of type assertions everywhere).

This change removes some vestigial arbitrary type support.

magik6k (approved):

PR #127, opened 2019-04-09
fix the keytransform datastore's query implementation

Stebalien:
The namespace Datastore shouldn't have to touch the mess with the query. It was also only partially implementing the query logic, passing through filters, etc. without applying the key transform.

Also adds some tests.

### DIFF
--- a/keytransform/keytransform.go
+++ b/keytransform/keytransform.go
@@ -5,60 +5,62 @@ import (
 	dsq "github.com/ipfs/go-datastore/query"
 )
 
-type Pair struct {
-	Convert KeyMapping
-	Invert  KeyMapping
+// Wrap wraps a given datastore with a KeyTransform function.
+// The resulting wrapped datastore will use the transform on all Datastore
+// operations.
+func Wrap(child ds.Datastore, t KeyTransform) *Datastore {
+	if t == nil {
+		panic("t (KeyTransform) is nil")
+	}
+
+	if child == nil {
+		panic("child (ds.Datastore) is nil")
+	}
+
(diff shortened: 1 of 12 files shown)


keytransform/keytransform.go
```diff
@@ -78,9 +80,101 @@ func (d *Datastore) Query(q dsq.Query) (dsq.Results, error) {
...
+	for i, o := range child.Orders {
+		switch o.(type) {
+		case dsq.OrderByValue, *dsq.OrderByValue,
+			dsq.OrderByValueDescending, *dsq.OrderByValueDescending:
+			// Key doesn't matter.
+			continue
+		case dsq.OrderByKey, *dsq.OrderByKey,
+			dsq.OrderByKeyDescending, *dsq.OrderByKeyDescending:
+			if orderPreserving {
+				child.Orders = child.Orders[:i+1]
```
magik6k, line 126: I feel like I'm missing something here. Why drop other orders here?

magik6k (commented):
Other than that 1 question LGTM

keytransform/keytransform.go
Stebalien: We sort hierarchically. That is, when comparing two entries, we compare with the first "order", then the second order if equal, then the third order if still equal, etc. Given that keys are unique, "order by key" will never return "equal" so there's no point in applying further orders.

Stebalien: I've added a comment and improved the test coverage more to get codecov to stop complaining.

magik6k (approved):

PR #128, opened 2019-04-13
autobatch: batch deletes

Stebalien:
part of fixing: https://github.com/libp2p/go-libp2p-kad-dht/issues/316#issuecomment-482753415


vyzo (approved):
LGTM. We might want to have an autoflush option.

Stebalien:
> We might want to have an autoflush option.

(I just made it flush the same way put does; should have done that in the first place)

PR #129, opened 2019-04-17
sync: apply entire query while locked

Stebalien:
This isn't a problem right _now_ (because we only have one non-sync datastore, MapDatastore) but it still makes me nervous.

### DIFF
--- a/sync/sync.go
+++ b/sync/sync.go
@@ -61,11 +61,28 @@ func (d *MutexDatastore) Delete(key ds.Key) (err error) {
 	return d.child.Delete(key)
 }
 
-// KeyList implements Datastore.KeyList
+// Query implements Datastore.Query
 func (d *MutexDatastore) Query(q dsq.Query) (dsq.Results, error) {
 	d.RLock()
 	defer d.RUnlock()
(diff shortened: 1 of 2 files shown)


sync/sync.go
```diff
@@ -61,11 +61,26 @@ func (d *MutexDatastore) Delete(key ds.Key) (err error) {
...
+	results, err := d.child.Query(q)
+	if err != nil {
+		return nil, err
+	}
+
+	// Apply the entire query while locked. Non-sync datastores may not
+	// allow concurrent queries.
+	var res []dsq.Result
+	for {
+		next, ok := results.NextSync()
```
magik6k: Why not just use `results.Rest()` and pass the resulting `[]Entries` to `ResultsWithEntries`? To pass errors exactly?

magik6k (commented):
1 question, other than that LGTM.

sync/sync.go
Stebalien: I wanted to be able to return a bunch of results and then an error. However, in retrospect, that's probably not worth it.

magik6k (approved):

PR #132, opened 2019-08-21
feat: make delete idempotent

### LINKED ISSUE #104: Define Delete to be idempotent
Moving from: https://github.com/ipfs/go-ds-flatfs/issues/30

1. We usually only care if the operation succeeded.
2. We *definitely* don't want to abort a batch operation because a delete failed.
3. User code often just aborts/returns errors without thinking about them.
4. Not all datastores *can* tell us if there was nothing to delete (e.g., badger).
5. Flatfs no longer reliably tells us this.
6. HTTP delete is defined to be idempotent so many web-backed datastores won't be able to fulfill this contract either. (also, there's a *reason* HTTP delete is idempotent...).
7. 

This boils down to:

1. We can't reliably return ErrNotFound on delete making it useless except, maybe, as a source of metrics.
2. Having to handle ErrNotFound on delete in user code is annoying.
3. Having to *try* to return ErrNotFound on delete is annoying at best.

---

Proposal: Define it to be idempotent (i.e., return `nil` when the key isn't found).

Stebalien:
fixes #104

### DIFF
--- a/autobatch/autobatch.go
+++ b/autobatch/autobatch.go
@@ -75,10 +75,6 @@ func (d *Datastore) Flush() error {
 		var err error
 		if o.delete {
 			err = b.Delete(k)
-			if err == ds.ErrNotFound {
-				// Ignore these, let delete be idempotent.
-				err = nil
-			}
 		} else {
 			err = b.Put(k, o.value)
 		}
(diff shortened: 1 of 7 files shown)


Stebalien:
Let's not hold up a _bugfix_ on feature bike-shedding.

dirkmc (approved):
LGTM 👍

PR #134, opened 2019-09-06
add a Size field to Query's Result

MichaelMure:
Rational is that some datastore can easilty get the size of a value when listing only keys, so we might as well have a way to pass that. One way this could be useful is when implementing a cache for Has/GetSize.

### DIFF
--- a/query/query.go
+++ b/query/query.go
@@ -66,7 +66,7 @@ type Query struct {
 	Offset            int      // skip given number of results
 	KeysOnly          bool     // return only keys.
 	ReturnExpirations bool     // return expirations (see TTLDatastore)
-	ReturnsSize       bool     // always return sizes. If not set, datastore impl can return
+	ReturnsSizes      bool     // always return sizes. If not set, datastore impl can return
 	//                         // it anyway if it doesn't involve a performance cost. If KeysOnly
 	//                         // is not set, Size should always be set.
 }
--- a/query/query.go
+++ b/query/query.go
@@ -66,6 +66,9 @@ type Query struct {
 	Offset            int      // skip given number of results
 	KeysOnly          bool     // return only keys.
 	ReturnExpirations bool     // return expirations (see TTLDatastore)
+	ReturnsSize       bool     // always return sizes. If not set, datastore impl can return
+	//                         // it anyway if it doesn't involve a performance cost. If KeysOnly
+	//                         // is not set, Size should always be set.
 }
 
 // String returns a string represenation of the Query for debugging/validation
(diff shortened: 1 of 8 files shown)


query/query.go
```diff
@@ -70,6 +70,7 @@ type Query struct {
 // Entry is a query result entry.
 type Entry struct {
 	Key        string    // cant be ds.Key because circular imports ...!!!
+	Size       int       // Might be zero if the datastore doesn't support listing the size with KeysOnly
```
Stebalien: So, a value can _also_ just be empty. We'd have to use -1 to mean "not set" (or a `HasSize` flag).

What if we:

1. Added a field to the query to _request_ the size (like we do with expirations).
2. Always included the size if requested, potentially doing extra work.

This is a breaking change for datastore implementations but it's a pretty minor breaking change.

Stebalien:
Hm. I just wanted this when messing with `ipfs repo stat` (I wanted to be able to report sizes for various record types).

query/query.go
MichaelMure: That works for me. A datastore layer can also hijack the query and enable the flag if it's needed (say, a caching layer) or considered free (say, S3).

MichaelMure:
I changed the semantic so that:
- the calling side can opt-in to always have the size set, even with `KeysOnly`, even it that implies a performance cost
- a datastore implementation can optionally always return the size if it doesn't imply a performance cost

I feel like that the best compromise. What do you think ?

Stebalien (approved):
LGTM!

Stebalien:
Could you file pull requests against the other datastores before we merge this?

* [x] go-ds-badger
* [ ] ~go-ds-bolt~ broken
* [x] go-ds-flatfs
* [x] go-ds-leveldb
* [x] go-ds-redis
* [x] go-ds-s3

Stebalien (changes requested):
Actually, this needs to update the mount datastore. We don't need to forward the `RequestSize` to child queries.

While we're at it, we should add tests to `test/suite.go`.

MichaelMure:
> Actually, this needs to update the mount datastore. We don't need to forward the RequestSize to child queries.

Not sure I follow, It doesn't at the moment, but it should, right ?

Stebalien:
Ah, sorry, yes, we _need_ to forward ReturnsSize.

Also, I noticed the variable is currently `ReturnsSize`, it should be `ReturnSizes` (we're _asking_ the datastore to return the sizes).

MichaelMure:
- fixed the mount datastore
- renamed ReturnsSize to ReturnsSizes
- added a test for ReturnsSizes

Should be good to go now

MichaelMure:
> Could you file pull requests against the other datastores before we merge this?
> 
* [X]  go-ds-badger : issue opened, https://github.com/ipfs/go-ds-badger/pull/76
* [ ]  go-ds-bolt : can't file issue there, first step: https://github.com/ipfs/go-ds-bolt/pull/4
* [X]  go-ds-flatfs : issue opened, https://github.com/ipfs/go-ds-flatfs/pull/59
* [X]  go-ds-leveldb : issue opened, https://github.com/ipfs/go-ds-leveldb/pull/35
* [X]  go-ds-redis : issue opened, but also **doesn't support query to begin with**
* [X]  go-ds-s3: https://github.com/ipfs/go-ds-s3/pull/45

Stebalien:
Unfortunately, I can't merge this until we actually have PRs for those issues. Otherwise, all those datastores will return invalid results.

MichaelMure:
Any chance the respective maintainers could help here ?

Stebalien:
Unfortunately, I am the de facto maintainer.

MichaelMure:
I did a pass on all of them but `go-ds-bolt` is problematic as it's broken at the moment.

Stebalien (approved):

Stebalien:
All merged and updated.

Issue #137, opened 2019-10-30
Asynchronous Datastores


aschmahmann:
## Proposal

1. Add a `Sync(prefix Key)` function to the `Datastore` interface.
    * This function will be a no-op when the datastore is in synchronous mode (the default).
    * Otherwise, `Sync(prefix)` guarantees that any `Put(prefix + ..., value)` calls that returned _before_ `Sync(prefix)` was called will be observed _after_ `Sync(prefix)` returns, even if the program crashes.
2. Insert calls to `Sync` where appropriate (in go-ipfs and go-libp2p).
3. When ready, turn off sync writes in _go-ipfs's_ datastore (by default). (we'll have an experimental transition with heavy testing)

Notes:

1. We're not changing the default behavior. Datastores will still write synchronously unless configured not to do so.
2. Put will either completely put a value or not put a value. Even when sync writes is turned off, the datastore will never be left in a corrupt state.

## Motivation

Writing to disk synchronously has poor performance and is rarely necessary.

Poor performance: `ipfs add` performance is _doubled_ (on linux/ext4) when badger is used and synchronous writes are turned off.

Rarely necessary:
* The DHT expects some number of nodes to be faulty so losing a few records is usually fine.
* IPFS only guarantees that blocks are persisted when pinned. There's no reason to sync after every write.
  * Note: For now, we'll likely want to explicitly `sync` after a full `ipfs add` as most users have GC turned off and expect the data to be persisted anyways. However, doing this _once_ is cheaper than doing it for every write.
* The peerstore _definitely_ doesn't need synchronous writes.

## Alternatives

* Create a buffered/batching/async wrapper. This is what go-ipfs _currently_ does but we could do better.
* Use the "autobatching" datastore.

However:

1. Buffering/caching isn't easy.
2. Unlike buffering inside the OS, they can't (easily) respond to memory pressure.
3. Conversely, they force one to _eagerly_ sync/flush periodically instead of as-needed. The OS knows when we have enough memory to keep writing into memory.

@Stebalien @whyrusleeping @raulk Seem like a reasonable plan?

aschmahmann:
It's looking a lot like Badger on Windows **only** supports asynchronous writes due to issues with Golang not respecting Windows permissions.

This means that until there's a fix in Golang/Badger then if we want to support Badger we should support asynchronous datastores.

whyrusleeping:
[...]

PR #140, opened 2019-11-26
Support Asynchronous Writing Datastores

aschmahmann:
Per #137 we would like to support Datastores that internally have asynchronous writes. We are doing this by adding a `Sync(prefix ds.Key)` function to the Datastore interface.

The plan is:
1) Get this PR reviewed for a sanity check
2) Get the datastore PRs submitted and reviewed and ready to be accepted (their go.mod files will reference this PR's branch)
   - [x] https://github.com/ipfs/go-ds-measure/pull/23
   -  [x] https://github.com/ipfs/go-ds-s3/pull/46
   -  [x] https://github.com/ipfs/go-ds-redis/pull/6
   -  [x] https://github.com/ipfs/go-ds-leveldb/pull/36
   -  [x] https://github.com/ipfs/go-ds-flatfs/pull/60
   -  [x] https://github.com/ipfs/go-ds-badger/pull/77
   - Note: If some of the datastores are proving tricky to write a Sync command for we may not block for them all to be approved before we continue.
3) Merge this PR
4) Update and merge the other PRs
5) Update some areas of go-ipfs to use the `Sync()` command on their datastores
6) Switch the default datastore in go-ipfs to be an asynchronous one (i.e. Badger with Async writes enabled)

### DIFF
--- a/examples/fs.go
+++ b/examples/fs.go
@@ -63,6 +63,12 @@ func (d *Datastore) Put(key ds.Key, value []byte) (err error) {
 	return ioutil.WriteFile(fn, value, 0666)
 }
 
+// Sync would ensure that any previous Puts under the prefix are written to disk.
+// However, they already are.
+func (d *Datastore) Sync(prefix ds.Key) error {
+	return nil
+}
+
 // Get returns the value for given key
 func (d *Datastore) Get(key ds.Key) (value []byte, err error) {
 	fn := d.KeyFilename(key)
--- a/examples/fs.go
+++ b/examples/fs.go
@@ -103,8 +103,9 @@ func (d *Datastore) Query(q query.Query) (query.Results, error) {
 
 	walkFn := func(path string, info os.FileInfo, err error) error {
 		// remove ds path prefix
-		if strings.HasPrefix(path, d.path) {
-			path = path[len(d.path):]
+		relPath, err := filepath.Rel(d.path, path)
+		if err == nil {
+			path = filepath.ToSlash(relPath)
 		}
 
 		if !info.IsDir() {
@@ -167,7 +168,7 @@ func (d *Datastore) DiskUsage() (uint64, error) {
 			log.Println(err)
 			return err
 		}
-		if f != nil {
+		if f != nil && f.Mode().IsRegular() {
 			du += uint64(f.Size())
 		}
 		return nil
--- a/mount/mount.go
+++ b/mount/mount.go
@@ -189,6 +189,24 @@ func (d *Datastore) Put(key ds.Key, value []byte) error {
 	return cds.Put(k, value)
 }
 
+// Sync implements Datastore.Sync
+func (d *Datastore) Sync(prefix ds.Key) error {
+	// Sync all mount points below the prefix
+	// Sync the mount point right at (or above) the prefix
+	dstores, mountPts, rest := d.lookupAll(prefix)
+	for i, suffix := range rest {
+		if err := dstores[i].Sync(suffix); err != nil {
+			return err
+		}
+
+		if mountPts[i].Equal(prefix) || suffix.String() != "/" {
+			return nil
+		}
+	}
+
+	return nil
+}
+
 func (d *Datastore) Get(key ds.Key) (value []byte, err error) {
 	cds, _, k := d.lookup(key)
 	if cds == nil {
--- a/autobatch/autobatch.go
+++ b/autobatch/autobatch.go
@@ -64,6 +64,35 @@ func (d *Datastore) Put(k ds.Key, val []byte) error {
 	return nil
 }
 
+// Sync flushes all operations on keys at or under the prefix
+// from the current batch to the underlying datastore
+func (d *Datastore) Sync(prefix ds.Key) error {
+	b, err := d.child.Batch()
+	if err != nil {
+		return err
+	}
+
+	for k, o := range d.buffer {
+		if !(k.Equal(prefix) || k.IsDescendantOf(prefix)) {
+			continue
+		}
+
+		var err error
+		if o.delete {
+			err = b.Delete(k)
+		} else {
+			err = b.Put(k, o.value)
+		}
+		if err != nil {
+			return err
+		}
+
+		delete(d.buffer, k)
+	}
+
+	return b.Commit()
+}
+
 // Flush flushes the current batch to the underlying datastore.
 func (d *Datastore) Flush() error {
 	b, err := d.child.Batch()
--- a/autobatch/autobatch_test.go
+++ b/autobatch/autobatch_test.go
@@ -103,3 +103,149 @@ func TestFlushing(t *testing.T) {
 		t.Fatal("wrong value")
 	}
 }
+
+func TestSync(t *testing.T) {
+	child := ds.NewMapDatastore()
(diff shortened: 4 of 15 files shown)


examples/fs.go
```diff
@@ -103,8 +109,9 @@ func (d *Datastore) Query(q query.Query) (query.Results, error) {
 
 	walkFn := func(path string, info os.FileInfo, err error) error {
 		// remove ds path prefix
-		if strings.HasPrefix(path, d.path) {
-			path = path[len(d.path):]
+		relPath, err := filepath.Rel(d.path, path)
```
Stebalien, line 112: (this may break in weird cases (.., etc.). but this is just an example.

mount/mount.go
```diff
@@ -189,6 +189,15 @@ func (d *Datastore) Put(key ds.Key, value []byte) error {
 	return cds.Put(k, value)
 }
 
+// Sync implements Datastore.Sync
+func (d *Datastore) Sync(prefix ds.Key) error {
+	cds, _, k := d.lookup(prefix)
+	if cds == nil {
+		return ErrNoMount
+	}
+	return cds.Sync(k)
```
Stebalien: We need to sync _all_ datastores under the prefix. This will only sync the one responsible for the given key.

If we have two mounts `["/", "/blocks"]`, `Sync("/")` will currently _just_ sync "/".

Stebalien: Maybe we can modify the test datastore to record all `Sync` calls then use that o make sure sync is getting called correctly?

Stebalien (changes requested):
The mount datastore needs a small fix and a test but otherwise LGTM.

mount/mount.go
aschmahmann: ah cool, I didn't realize we wanted mount to be able to do that. I figured that if we wanted submounts that the "/foo" would point to another mount datastore that was "/foo/bar".

should be fixable though. will take a look later today

examples/fs.go
aschmahmann: Yep, I think for the example datastore not worth thinking too hard about. If similar issues come up in FlatFS we'll deal with it there.

mount/mount.go
Stebalien: Yeah, `lookup` is just for finding the datastore where we should read/write a key.

mount/mount.go
```diff
@@ -189,6 +189,42 @@ func (d *Datastore) Put(key ds.Key, value []byte) error {
...
+			if err := m.Datastore.Sync(ds.NewKey("/")); err != nil {
+				return err
+			}
+			atLeastOneSync = true
+		}
+	}
+
+	if !atLeastOneSync {
+		return ErrNoMount
+	}
```
Stebalien: I'd expect syncing a non-existent key to return success: there's nothing there so there's nothing to persist.

`Put` returns an error because the put fails. `Get` just returns `ErrNotFound`.

mount/mount.go
```diff
@@ -189,6 +189,42 @@ func (d *Datastore) Put(key ds.Key, value []byte) error {
 	return cds.Put(k, value)
 }
 
+// Sync implements Datastore.Sync
+func (d *Datastore) Sync(prefix ds.Key) error {
+	// Sync all mount points below the prefix
+	// Sync the mount point right at (or above) the prefix
+	// Return ErrNoMount if no mount points are synced
+	atLeastOneSync := false
+	for _, m := range d.mounts {
```
Stebalien: Can we use `lookupAll`?

Stebalien (changes requested):

bonedaddy:
After this PR is merged can we have a version bump according to semver practices since this is a breaking change to the `Datastore` interface?

Stebalien:
> After this PR is merged can we have a version bump according to semver practices since this is a breaking change to the Datastore interface?

We will bump the minor version because the package version is sub-zero.

bonedaddy:
> > After this PR is merged can we have a version bump according to semver practices since this is a breaking change to the Datastore interface?
> 
> We will bump the minor version because the package version is sub-zero.

Won't break `go get -u=patch` so :rocket:

mount/mount.go
aschmahmann: No problem here.

My thought was that if you tried to access an unmounted part of the hierarchy you might want to get an error. It's reasonable (and easier for testing purposes) to say that we did Sync all puts/deletes under that key and there happen to be none there because that key is unmounted.

mount/mount.go
aschmahmann: np, it might do a few more comparisons but it'll look much cleaner

aschmahmann: lmk how it looks now

mount/mount.go
```diff
@@ -189,6 +189,24 @@ func (d *Datastore) Put(key ds.Key, value []byte) error {
...
+	// Sync the mount point right at (or above) the prefix
+	dstores, mountPts, rest := d.lookupAll(prefix)
+	for i, suffix := range rest {
+		if err := dstores[i].Sync(suffix); err != nil {
+			return err
+		}
+
+		if mountPts[i].Equal(prefix) || suffix.String() != "/" {
+			return nil
+		}
```
Stebalien, line 204: Hm. This looks like a bug in `lookupAll`. `lookupAll` should only return mounts that can contain descendents of `key`.

Stebalien (approved):

mount/mount.go
Stebalien: We can merge this as-is and fix it later or move this logic into `lookupAll` now.

aschmahmann: Could you clarify what you think the bug is here? It looks like the documented behavior for `lookupAll`.

Stebalien: `lookupAll` is supposed to return all datastores that _might_ contain keys that are descendants of the given key. That means we should be calling `Sync` on all datastores returned by `lookupAll` datastores and we shouldn't have this extra check.

However, if we have `["/foo", "/foo/bar"]`, `"/foo"` should "cover" `"/foo/bar"` given that `lookup` will always return `"/foo"`. However, when given the prefix `"/foo"`, `lookupAll` will return _both_.

Note: this is only _technically_ a bug. Nobody would ever mount `"/foo"` over `"/foo/bar"`.

aschmahmann: > Nobody would ever mount "/foo" over "/foo/bar"

I thought it was expected behavior that we might mount "/" and also "/foo", or is the root a special case?

Stebalien: My first example was bad: order matters. If we mount "/" over "/foo", "/" will win (I think? check me here). In go-ipfs, we mount "/blocks" over "/".

Stebalien: I say "nobody would ever mount "/foo" over "/foo/bar" because "/foo" is a prefix of "/foo/bar" (and we check mounts in-order).

Stebalien:
LGTM but let's merge this tomorrow. Hitting merge then going to bed is a bad idea (I mean, it's one of my favorite bad ideas but I think it's time to call it a day).

mount/mount.go
Stebalien: I was wrong. We reverse-sort datastores so "/foo/bar" always comes before "/foo".

However, the bug was valid. Fix in #141.

PR #141, opened 2019-12-03
Only count a key as an ancestor if there is a separator

Stebalien:
Also make sure to clean and normalize keys before using them as prefixes.

BREAKING CHANGES:

* `myds.Query(Query{Prefix:"/foo"})` will no longer match "/foobar" (or even "/foo"). This is usually what the user expects, we had a tendency to normalize "/foo/" to "/foo" (when we clean keys), and many datastores can't efficiently search for prefixes that aren't on path-boundaries anyways.
* Given a mount datastore with mounts `["/foo", "/"]`, `myds.Put("/foo", "bar")` will put the value to the datastore mounted at "/", not "/foo", as the key "/" and "" usually doesn't make sense.

While technically breaking, these changes are much more likely to fix bugs than they are to break things.


key.go
```diff
@@ -212,20 +212,17 @@ func (k Key) ChildString(s string) Key {
 //   NewKey("/Comedy").IsAncestorOf("/Comedy/MontyPython")
 //   true
 func (k Key) IsAncestorOf(other Key) bool {
-	if other.string == k.string {
-		return false
-	}
-	return strings.HasPrefix(other.string, k.string)
+	// equivalent to HasPrefix(other, k.string + "/")
+	return len(other.string) > len(k.string) && // the child is longer
+		other.string[len(k.string)] == '/' && // the child has a separator in the right place
```
aschmahmann: The problem here is when `k.string == "/"` since `other.string[1]` should never be `/`.

I'll give this a try.

aschmahmann: any thoughts on what I can assume the state of a `Key` is? For example, do I need to run `Clean` to make sure that root isn't represented as `/./././././././`?

Stebalien: Keys are cleaned when they're created. You can assume they're clean.

mount/mount.go
```diff
@@ -159,21 +159,21 @@ func (h *querySet) next() (query.Result, bool) {
 // /aoe/      not matching
 func (d *Datastore) lookupAll(key ds.Key) (dst []ds.Datastore, mountpoint, rest []ds.Key) {
 	for _, m := range d.mounts {
-		p := m.Prefix.String()
-		if len(p) > 1 {
-			p = p + "/"
-		}
-
-		if strings.HasPrefix(p, key.String()) {
+		if m.Prefix.IsDescendantOf(key) {
```
Stebalien, line 209: I'm now using proper key functions.

mount/mount.go
```diff
@@ -159,21 +159,21 @@ func (h *querySet) next() (query.Result, bool) {
...
 			r := strings.TrimPrefix(key.String(), m.Prefix.String())
 
 			dst = append(dst, m.Datastore)
 			mountpoint = append(mountpoint, m.Prefix)
 			rest = append(rest, ds.NewKey(r))
+
+			// We've found an ancestor (or equal) key. We might have
+			// more general datastores, but they won't contain keys
+			// with this prefix so there's no point in searching them.
+			break
```
Stebalien, line 223: This is https://github.com/ipfs/go-datastore/pull/140#discussion_r352838087. Thinking back on this, I think it was a bug.

 Given the mounts `["/foo", "/"]`, `lookupAll("/foo/bar")` would return _both_ (there's now a test for this). However, it should have only returned "/foo" because no keys under "/foo/bar" would end up in "/".

mount/mount.go
```diff
@@ -191,15 +191,11 @@ func (d *Datastore) Put(key ds.Key, value []byte) error {
...
-	dstores, mountPts, rest := d.lookupAll(prefix)
+	dstores, _, rest := d.lookupAll(prefix)
 	for i, suffix := range rest {
 		if err := dstores[i].Sync(suffix); err != nil {
 			return err
 		}
-
-		if mountPts[i].Equal(prefix) || suffix.String() != "/" {
-			return nil
-		}
```
Stebalien, line 202: The fix in `lookupAll` should make this unnecessary.

mount/mount.go
Stebalien: Note: this may have been impacting the performance of GC.

query/query_impl.go
```diff
@@ -116,7 +118,23 @@ func NaiveOrder(qr Results, orders ...Order) Results {
...
+				prefix = "/" + prefix
+			}
+			prefix = path.Clean(prefix)
+		}
+		// If the prefix isn't "/", end it in a "/" so we only find keys
+		// _under_ the prefix.
+		if prefix != "/" {
+			prefix += "/"
+		}
+		qr = NaiveFilter(qr, FilterKeyPrefix{prefix})
```
Stebalien, line 137: We weren't cleaning/fixing keys.

Stebalien:
> myds.Query(Query{Prefix:"/foo"}) will no longer match "/foobar" (or even "/foo"). This is usually what the user expects, we had a tendency to normalize "/foo/" to "/foo" (when we clean keys), and many datastores can't efficiently search for prefixes that aren't on path-boundaries anyways.

Note: if we _don't_ do this, we need to remember to fix the key transform datastore.

Stebalien:
@whyrusleeping says

> uhhhh, i think normalization seems fine to me

whyrusleeping:
Yeah, seems fine to me. We can add more docs around our key model to compensate for the potential weirdness

query/query_test.go
```diff
@@ -88,14 +88,12 @@ func TestNaiveQueryApply(t *testing.T) {
 
 	q = Query{
 		Limit:  3,
```
aschmahmann: Limit isn't being tested anymore since we only have one result

mount/mount.go
aschmahmann: Are the comments above this function still accurate? What should happen with `/ao/e/` here?

mount/mount.go
aschmahmann: seems reasonable, this is a contract change but if it increases performance/reliability and no one is utilizing the obscure case then we may as well go for it.

Are we pretty confident that we're not doing queries for things like `/blocks` even though `/blocks` is mounted over `/`?

mount/mount.go
```diff
@@ -39,7 +39,7 @@ var _ ds.Datastore = (*Datastore)(nil)
 
 func (d *Datastore) lookup(key ds.Key) (ds.Datastore, ds.Key, ds.Key) {
 	for _, m := range d.mounts {
-		if m.Prefix.Equal(key) || m.Prefix.IsAncestorOf(key) {
+		if m.Prefix.IsAncestorOf(key) {
```
aschmahmann, line 73: To clarify, is the intent here that if I've mounted `["/", "/foo"]` and then I do a put to `/foo` that the key will end up in the `/` mount instead of the `/foo` mount?

Is there any reason to support this or would we be safer just throwing an error instead?

aschmahmann (approved):
Overall LGTM. Left a few comments and requests for clarification

mount/mount.go
Stebalien: I'm not really sure what that comment is trying to say (what's A, B?). It would return the /ao/e/ mount point, and / as "rest". I've fixed the documentation and added a test.

mount/mount.go
Stebalien: We are, but that should be fine. That is, we're querying for blocks in /blocks, not for blocks in /. As of this change, `lookupAll("/blocks")`, should return only the datastore mounted at /blocks.

query/query_test.go
Stebalien: I've added some more keys to improve the tests.

PR #144, opened 2019-12-15
fix go-check path to use "gopkg.in/check.v1"

xiegeo:
Solves the problem of `check` being required by two different paths, which creates errors when running `go test all`.

Also, updates check to the latest version, and fix 3 spelling mistakes fixed by `misspell`.


Stebalien (approved):
Thanks!

PR #151, opened 2020-02-15
test suite: Add ElemCount to control how many elements are added.

hsanjuan:
Context: for every subtestQuery the suite Puts 400 elements to the datastore. In the case of go-ds-crdt, every Put is actually about 4 insertions. The final following 400 deletions made during cleanup are actually 400 more insertions (to the Tombstone set). Multiply that for every Subtest and for all the Permutations.

Additionally every test writes the same set of keys over an over, which means that the number of crdt blocks related to every key grows. Every go-ds-crdt Has and Get operations needs to loop over those and triggers related subqueries.

Finally the map datastore is not optimized for prefixed queries so every Query() will loop-and-match ALL the things in the database.

The result is that it takes very long to run these tests for go-ds-crdt. Reducing ElemCount helps significantly.

### DIFF
--- a/test/basic_tests.go
+++ b/test/basic_tests.go
@@ -12,6 +12,17 @@ import (
 	dsq "github.com/ipfs/go-datastore/query"
 )
 
+// ElemCount sets with how many elements the datastore suit
+// tests are usually run with. Best to set to round numbers like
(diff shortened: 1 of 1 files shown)


Stebalien (approved):
SGTM.

PR #152, opened 2020-02-18
/test: fix bad ElemCount/10 lenght (should not be divided)

hsanjuan:
I had changed this manually and then missed it.


Stebalien (approved):

Stebalien:
Oops. I should have caught that on review.

PR #157, opened 2020-04-27
support flatfs fuzzing

willscott:
I'd forgotten that flatfs isn't a transactional implementation, but probably an important one to test.
This change allows any datastore to get basic fuzzing, and opts up to transaction testing when the implementation meets that interface.


willscott:
merging as minor follow-on to previous setup / unblocking.

ribasushi:
Sorry, stretched too thin... :(

PR #158, opened 2020-04-30
Add the generated flatfs stub, since it cannot be auto-generated

willscott:


willscott:
essentially, this is a file that was forgotten as part of #157

willscott:
merging so that master is ready when being reviewed in the context of https://github.com/google/oss-fuzz/pull/3739

PR #162, opened 2020-08-15
Add test to ensure that Delete returns no error for missing keys

ncabatoff:
Make sure that Delete does not treat a missing key as an error.  Would've caught https://github.com/ipfs/go-ds-crdt/issues/62.


Stebalien (approved):
Thanks. For context, this API was recently (ish) changed. That's why go-ds-crdt returned an error.

PR #180, opened 2021-10-13
autobatch: thread-safe, debounce, max delay and implement Batching

MichaelMure:
Revamp the venerable autobatch in different ways:
- make it thread safe (test -race is happy)
- deduplicate writes (avoid writing the same key multiple time), which make it act as a debounce
- introduce a maximum delay before the write happen, to avoid keeping the pending writes in memory if the count trigger is not reached
- implement the Batching interface for compatibility with more usecases


aschmahmann:
@whyrusleeping you might be interested in this.

autobatch/autobatch.go
```diff
@@ -23,28 +31,79 @@ type op struct {
...
-// of the memory pool is given by size.
-func NewAutoBatching(d ds.Batching, size int) *Datastore {
-	return &Datastore{
-		child:            d,
-		buffer:           make(map[ds.Key]op, size),
-		maxBufferEntries: size,
+// batches writes using the given Batching datastore. The maximum number of
+// write before triggering a batch is given by maxWrite. The maximum delay
+// before triggering a batch is given by maxDelay.
+func NewAutoBatching(child ds.Batching, maxWrite int, maxDelay time.Duration) *Datastore {
```
MichaelMure, line 39: Strictly speaking this does break compatibility, although it's an easy change and I'm not sure many people use it.
It would be possible to maintain the previous signature with `maxDelay` set to infinity and have a second constructor with the new parameter.

autobatch/autobatch.go
```diff
@@ -23,28 +31,79 @@ type op struct {
...
+		timer = nil
+
+		b, err := d.prepareBatch(nil)
+		if err != nil {
+			log.Println(err)
+			return
+		}
+		err = b.Commit()
+		if err != nil {
+			log.Println(err)
```
guseggert: Would it make sense to use https://github.com/ipfs/go-log here?

autobatch/autobatch.go
```diff
@@ -23,28 +31,79 @@ type op struct {
...
+	d.buffer[key] = op
+	d.mu.Unlock()
+	d.newWrite <- struct{}{}
+}
+
+func (d *Datastore) runBatcher() {
+	var timer <-chan time.Time
+
+	write := func() {
+		timer = nil
```
guseggert, line 63: This leaks the timer until it fires, e.g. if there is a high rate of ops that hit the maxWrite threshold really fast, we could end up with a lot of concurrent no-op timers running. I think it'd be better to use a proper timer and stop it here.

autobatch/autobatch.go
```diff
@@ -164,5 +236,32 @@ func (d *Datastore) Close() error {
 	if err2 != nil {
 		return err2
 	}
+	close(d.exit)
+	close(d.newWrite)
```
guseggert, line 242: I think this might be racy if the timer fires and prepares a batch, then this flushes and closes the underlying datastore, then the async writer tries to commit the batch but the underlying DS is now closed

Also this could return before the async writer finishes flushing things, which is surprising.

autobatch/autobatch_test.go
```diff
@@ -70,6 +72,9 @@ func TestFlushing(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	// flushing is async so we can rely on having it happening immediately
+	time.Sleep(100 * time.Millisecond)
```
guseggert, line 76: can we make these deterministic to avoid flaky tests?  (stub out time fns, inject mock clock. etc)

guseggert (commented):
I can't comment on Query(), so will comment here.  Is the intention of flushing in Query() before issuing the query, to make sure that there are no pending ops before sending the query? If so, then I think that's racy (it unlocks after flushing and re-locks before querying).

autobatch/autobatch_test.go
MichaelMure: Is there a library you would recommend for that?

autobatch/autobatch.go
MichaelMure: Fixed in a new commit, feel free to squash before merging if you feel like it.

MichaelMure:
> I can't comment on Query(), so will comment here. Is the intention of flushing in Query() before issuing the query, to make sure that there are no pending ops before sending the query? If so, then I think that's racy (it unlocks after flushing and re-locks before querying).

This is correct, but at the same time Query can be *very* long lived (I used some hours-long Query before) so I don't think blocking writes during a Query is reasonable.

autobatch/autobatch_test.go
guseggert: Recommend https://github.com/benbjohnson/clock

guseggert:
> > I can't comment on Query(), so will comment here. Is the intention of flushing in Query() before issuing the query, to make sure that there are no pending ops before sending the query? If so, then I think that's racy (it unlocks after flushing and re-locks before querying).
> 
> This is correct, but at the same time Query can be _very_ long lived (I used some hours-long Query before) so I don't think blocking writes during a Query is reasonable.

Ah that makes sense, thanks :)

autobatch/autobatch.go
guseggert: Would it make sense to use varargs functional opts here to a) preserve backwards compat and b) give us a way to extend this in the future?  (I think backwards compat would also require disabling the async flushing if maxDelay=0)

MichaelMure: That can be done but that would be a bit weird to keep maxWrite as a parameter and maxDelay as a functional arg.
Also, not sure how much time I can find to work on this.

BigLep:
@MichaelMure : I've converted this to a draft.  If/once you've incorporated feedback, feel free to publish and we'll happily take a look.

BigLep:
@MichaelMure : are you going to take this, or should we close this for now?

MichaelMure:
@BigLep I'm sure there is value in there but I ended up not using any of this in my project as things took a different turn. I suppose we can close.

PR #181, opened 2021-10-26
feat: add context to interfaces

guseggert:
This adds contexts to all the Datastore interfaces. The motivation for
this change is for instrumentation, not cancellation, although these
can certainly be used in the future for adding cancellation. We
default to adding context to everything, even if we don't immediately
use it, because we might need them in the future and making this
change again is quite painful due to the large number of repos this
fans out to.

Note that we have not added context to Close() methods, due to it
being surprising given that it breaks the io.Closer interface, and
many Close() methods are quick and don't do much work.

This also disables the fuzz test, because it has a submodule which
transitively depends on this module, so it will fail to build until
this change is plumbed through go-ds-flatfs.


aschmahmann (approved):
Thanks. LGTM

burdiyan:
Just updated go-datastore in a project I'm working on, and was so surprised that everything blew up between supposedly minor version change. I do support adding a context, but maybe a better way could have been by adding separate methods for each instead of breaking everything.

lanzafame:
@aschmahmann Why was this not a v2 release?

PR #185, opened 2021-11-29
Release v0.5.1

aschmahmann:


guseggert (approved):

aschmahmann:
Note: We are releasing as v0.5.1, because the breaking change here should actually have been part of v0.5.0.

PR #187, opened 2022-01-06
Add LRUStore datastore wrapper

gammazero:
Package `lrustore` provides a datastore wrapper that limits the number of items stored to the specified capacity. When at capacity, the oldest items are removed in LRU to make root for new items. Putting and getting items maintains LRU order.

This is useful for implementing an LRU-cache on top of a datastore to maintain limited storage size, or to have an LRU-cache for values that may be too large to store in memory.

Note that while the keys and values are persisted in the datastore, the LRU-ordered list of keys is kept only in memory. Keeping the keys in memory only was done for performance so as not to maintain a linked list in the datastore.  This has two consequences:
- LRU order not maintained across restart
- Cannot have arbitrarily large LRU size

Reviewers, please evaluate whether the above is preferable, or if the LRU-ordered list of keys also needs to be persisted.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,7 @@
 module github.com/ipfs/go-datastore
 
 require (
+	github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da
 	github.com/google/uuid v1.1.1
 	github.com/ipfs/go-detect-race v0.0.1
 	github.com/ipfs/go-ipfs-delay v0.0.0-20181109222059-70721b86a9a8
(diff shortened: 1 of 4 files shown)


BigLep:
@gammazero : this isn't needed currently and it's not being needed right now, right?  Assuming that's the case, we don't want to add it here for now to reduce maintenance.  Do you want to stick it somewhere else for now and if it gets usage it moves over here later. Experimental/less maintained datastores can live in ipfs-shipyard or in the repo where they're used.

gammazero:
I will keep this somewhere else.  It was previously being used, but now is not, so agree that its better to have less to maintain.

Issue #160, opened 2020-05-07
Keytransform Only Supports Batched Datastores


bonedaddy:
I'm working on a codebase that currently uses a `datastore.Batching` type backed by `leveldb` using a namespace wrap. Unfortunately it appears that the namespace only supports `datastore.Batching` and can't be used in place of a `datastore.TxnDatastore`.

Is this intentional?

Stebalien:
We should probably be doing what we do in the mount datastore. There, we type assert to `datastore.Batching` on demand and return `ds.ErrBatchUnsupported` if batching is unsupported.

Issue #88, opened 2018-06-07
Find a better type structure


Stebalien:
Currently, we have:

1. A ton of extension interfaces: CheckedDatastore, ScrubbedDatastore, Batching, etc...
2. A ton of wrappers: Delayed, Mount, Sync, Autobatch, etc...

And unfortunately, every wrapper needs to implement and forward every extension interface to the underlying datastore. It's kind of a nightmare.

According to @bigs, however,

> Typeclasses and modules compose pretty cleanly stuck_out_tongue! Monad transformers exhibit that kind of extensionality pretty well!

I'm not sure about languages like Haskell but I'm pretty sure this can't be done in most languages. However, it's worth discussing as it would be great to have a better system.

Basically, what I want is (pseudo-rust):

```
trait Datastore { ... }
trait GCDatastore: Datastore { ... }

struct MyDs {
   ...
}

impl Datastore for MyDs { ... }

struct Delayed<D: Datastore> {
  inner: D,
  delay: Duration,
}

impl<D: Datastore> Datastore for Delayed<D> {}

// Can't express this:
forall<X> auto impl X for Delayed<D> where D: X via self.inner;

// That is, auto implement all traits implemented by D on Delayed<D> by proxying to self.inner.
```

This kind of behavior is usually achieved via the `Deref` trait. In rust, if a type `X` implements `Deref<Target=SomeOtherType>` and someone calls `some_x.foo()`, the compiler will first try to lookup `foo` on `X`, then on `SomeOtherType`, recursively (if `SomeOtherType` also implements `Deref`). However, this doesn't mean that the type `X` actually *implements* any of the traits implemented by `SomeOtherType`.

Thoughts?

Stebalien:
One solution is to define new types and use go composition:

```go

type Datastore interface {...}
type GCDatastore interface {...}

type MyDs struct {...}
// implement Datastore and GCDatastore for Delayed

type Delayed struct {
  Datastore Datastore
  Delay Duration
}

// Only implement Datastore for Delayed

...

// Compose them:
struct{Delayed, MyDs}{Delayed{Datastore, someDelay}, MyDs}
```

bigs:
Yeah, this seems pretty solid. I just kind of wonder if it breaks the facade of our interface altogether... Without generics you're going to have to have a direct reference to the underlying struct unless you're willing to resort to type assignment, which I think is probably OK. Like anything, it will have to be wielded responsibly.

Stebalien:
[...]

PR #188, opened 2022-03-23
feat: add Features + datastore scoping

guseggert:
This is ripped out of https://github.com/guseggert/go-ds-grpc, as I realized it solves a more general problem.

### Commit
The motivation for this is to enable "dispatching" datastores that
dynamically implement the type of the datastore they are dispatching
to, so that type assertions behave equivalently on the dispatcher
as on the dispatchee. We also want this to be backwards-compatible
with existing code using type assertions.

At a high level, this works by generating a concrete implementation of
every possible combination of "features", and then picking the right
implementation at runtime. This is necessary due to language
constraints in Go--it is currently impossible to create a concrete
type dynamically with reflection that implements an interface.

"Features" are introduced here, which are supplemental, optional
interfaces that datastores may implement. These are
backwards-compatible with existing "features", which are:

* Batching
* CheckedDatastore
* GCDatastore
* PersistentDatastore
* ScrubbedDatastore
* TTLDatastore
* TxnDatastore

New features can also be added in a backwards-compatible way. E.g. if
datastore A is scoped down to datastore B, a new feature F is added, and
then implemented on B, then A will continue to implement the same set
of features since it hasn't implemented F yet (and vice versa if F is
implemented on A but not B).

Examples of things this enables:

* Allow us to deprecate ErrBatchUnsupported (maybe...depending on mount datastores)
* Allow existing dispatching datastores to support all
features (keytransform, retrystore, MutexDatastore, autobatch, etc.)
* Features supported by a Mount datastore could be scoped down to the
intersection of all children
* Communication with data about what functionality a datastore
supports (e.g. for cross-language/RPC support)

Some related issues:

* https://github.com/ipfs/go-datastore/issues/160
* https://github.com/ipfs/go-datastore/issues/88


guseggert:
@aschmahmann pointed out yesterday that we could also use this, along with a new method on mount datastores to return a scoped datastore for a given key/key prefix/query, which effectively solves the problem with mixing multiple datastore types together with a mount datastore.

features.go
```diff
@@ -0,0 +1,142 @@
...
+	feat, known := featuresByName[name]
+	return feat, known
+}
+
+// FeaturesForDatastore returns the features supported by the given datastore.
+func FeaturesForDatastore(dstore Datastore) (features []Feature) {
+	if dstore == nil {
+		return nil
+	}
+	dstoreType := reflect.ValueOf(dstore).Type()
```
Jorropo:
```suggestion
	dstoreType := reflect.TypeOf(dstore)
```

features.go
```diff
@@ -0,0 +1,142 @@
...
+
+// FeaturesByNames returns the features with the given names, if they are known.
+func FeaturesByNames(names ...string) (features []Feature) {
+	for _, n := range names {
+		if feat, ok := featuresByName[n]; ok {
+			features = append(features, feat)
+		}
+	}
+	return
+}
```
Jorropo: Is this really necessary ? I always felt in go you just write the loop instead of biting the syntax sugar candy.

This function isn't more optimized or clearer than a loop calling `FeatureByName` and is non obvious about what it's failing behaviour is (what if the name isn't found ?),
me being a "fail loudly" kind of of person would expect it returning an error (which it don't), so a panic instead. But it silently swallow the error instread.

features_test.go
```diff
@@ -0,0 +1,80 @@
...
+		feats[0].Interface != (*BatchingFeature)(nil) ||
+		feats[0].DatastoreInterface != (*Batching)(nil) {
+		t.Fatalf("expected a batching feature, got %v", feats)
+	}
+
+	feats = FeaturesByNames(FeatureNameBatching, "UnknownFeature")
+	if len(feats) != 1 || feats[0].Name != FeatureNameBatching {
+		t.Fatalf("expected a batching feature, got %v", feats)
+	}
+}
```
Jorropo: This test is testing that you properly implemented a for loop, an if and an append.

I think this shows how silly it is to have a function dedicated to that in the first place.

null_ds.go
```diff
@@ -0,0 +1,120 @@
...
+
+import (
+	"context"
+
+	dsq "github.com/ipfs/go-datastore/query"
+)
+
+// NullDatastore stores nothing, but conforms to the API.
+// Useful to test with.
+type NullDatastore struct {
```
Jorropo, line 11: Self notes for later:
There is no reason for thoses types to be public, only the factory should be and we should be returning interfaces.

scoped/scoped.go
```diff
@@ -0,0 +1,42 @@
...
+	dstoreFeatures := getFeatureSet(dstore)
+
+	dstoreFeatureSet := map[string]bool{}
+	for _, f := range features {
+		if _, ok := dstoreFeatures[f.Interface]; ok {
+			dstoreFeatureSet[f.Name] = true
+		}
+	}
+
+	ctor := 0
```
Jorropo: Bit fields should be uints.

scoped/scoped.go
```diff
@@ -0,0 +1,42 @@
...
+	for i, f := range ds.Features() {
+		if _, ok := dstoreFeatureSet[f.Name]; ok {
+			ctor += (1 << i)
```
Jorropo:
```suggestion
			ctor |= (1 << uint(i))
```

scoped/scoped.go
```diff
@@ -0,0 +1,42 @@
...
+func getFeatureSet(dstore ds.Datastore) map[interface{}]bool {
+	features := ds.FeaturesForDatastore(dstore)
+	featureSet := map[interface{}]bool{}
+	for _, f := range features {
+		featureSet[f.Interface] = true
+	}
+	return featureSet
+}
+
+// Wrap returns a datastore based on the source, whose concrete type is scoped down to only the features supported by the target.
```
Jorropo, line 16: This sentence is wrong, as far as I can tell, this actually return the intersection of features.

I think the intersection is less useful, and I would like it to return an error instead of trying to silently continue.

scoped/impls.go
```diff
@@ -0,0 +1,2570 @@
+// Code generated by go generate; DO NOT EDIT.
+
+package scoped
```
Jorropo: pls rename that file `impls.gen.go` so we follow what we started in the IPv6 to ASN package.

Jorropo (changes requested):

Jorropo:
Minor changes should be needed.

I don't like routines that try to make up with degraded state silently.

If something is not happening as it should, fail loudly, so I (the programmer) can learn about it and fix it.

scoped/scoped.go
guseggert: The elements of set A that are also in set B is the definition of set intersection, the two statements are equivalent.

The reason this doesn't return an error if B is not a subset of A is to allow for both datastores to evolve independently, without breaking consumers of this.

features.go
guseggert: Yeah let me remove this, it is really a vestige from an earlier impl and no longer makes sense

scoped/scoped.go
Jorropo: For me this sentence assume that set A always is >= than set B. It's not really important.

Jorropo (approved):
LGTM

PR #192, opened 2022-06-06
New Feature: Add query filter by key substring

pyropy:
- Add new filter to enable query by key substring

### DIFF
--- a/query/filter.go
+++ b/query/filter.go
@@ -100,3 +100,15 @@ func (f FilterKeyPrefix) Filter(e Entry) bool {
 func (f FilterKeyPrefix) String() string {
 	return fmt.Sprintf("PREFIX(%q)", f.Prefix)
 }
+
+type FilterKeySubstring struct {
+	KeySubstring string
+}
+
+func (f FilterKeySubstring) Filter(e Entry) bool {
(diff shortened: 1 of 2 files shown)


guseggert:
Hi @pyropy, thanks for the PR. Can you explain your use case here? Also could you use this without adding it to go-datastore?

pyropy:
Hi @guseggert, I'd like to query complex keys by substring, for example let's say I have keys like this `/5/4/fe175c1f-25ff-455d-ae43-f21097e879c6/10:10`, `/5/4/fe175c1f-25ff-455d-ae43-f21097e879c6/9:10`, `/5/4/fe175c1f-25ff-455d-ae43-f21097e879c6/8:10` and I would like to query all items where key contains substring UUID `fe175c1f-25ff-455d-ae43-f21097e879c`.

This is my primary use-case and reason I've opened this PR. I would love to know if there is some other way to achieve this?

Jorropo:
The `query` package is slow already but the filters that exists could be made fast if we needed to (sorted indexes, binary search, other fancy db things, ...), contains just couldn't and will probably have `O(N)` complexity forever.

Given that everything you need is exported (the `Filter` interface, ...) this filter can live in different repos and be used just the same.

PR #193, opened 2022-08-17
feat: return error from txn.Discard()

guseggert:
To properly discard a transaction usually requires some external call
to the backing DB, which can fail. Currently implementations just
swallow any errors when discarding transactions, but this should be
surfaced to the application so that it can decide the appropriate
action (at the very least, log the error).

This is of course a breaking change and both datastore impls and
consumers will need to be updated.

### DIFF
--- a/datastore.go
+++ b/datastore.go
@@ -173,7 +173,7 @@ type Txn interface {
 	// them to the underlying Datastore. Any calls made to Discard after Commit
 	// has been successfully called will have no effect on the transaction and
 	// state of the Datastore, making it safe to defer.
-	Discard(ctx context.Context)
+	Discard(ctx context.Context) error
 }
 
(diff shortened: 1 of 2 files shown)


BigLep:
@guseggert : I put this on the 0.21 iteration.  Adjust if something different should be done.

Jorropo (approved):
:+1: Makes complete sense.
I would like to see PRs of a few implementations by Kubo linked to or from here before merging this.

I don't want to have this multirepo trap where we merge code that looks fine but is actually unusable without a bunch of more work that not ready yet.

guseggert:
Agreed this is a small change but has a big blast radius, which is why I didn't really prioritize it before. I'm going to downgrade this to draft until we estimate costs and then figure out if we really want this.

guseggert:
unlikely to do anything about this anytime soon, so closing

PR #197, opened 2022-11-20
add .github/workflows/cifuzz.yml

DavidKorczynski:
Add [CIFuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration/) workflow action to have fuzzers build and run on each PR.

This is a service offered by OSS-Fuzz where ipfs/go-datastore already runs (https://github.com/google/oss-fuzz/tree/master/projects/ipfs). CIFuzz can help detect regressions and catch fuzzing build issues early, and has a variety of features (see the URL above). In the current PR the fuzzers gets build on a pull request and will run for 300 seconds.

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,26 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Fuzzers
+      id: build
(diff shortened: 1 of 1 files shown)


aschmahmann:
@ipfs/ipdx should this be here, or in Unified CI (https://github.com/protocol/.github)?

galargh:
If it's a completely new workflow, I'd go with trying it out on a repo first and then proposing it for Unified CI.

PR #209, opened 2023-08-23
Add traced datastore

dennis-tra:
This pull request introduces the package `trace` that wraps a datastore and applies open telemetry tracing to all datastore interactions.

### DIFF
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -1,0 +1,228 @@
+// Package trace wraps a datastore where all datastore interactions are traced
+// with open telemetry.
+package trace
+
+import (
+	"context"
+	"fmt"
+	"io"
+
+	ds "github.com/ipfs/go-datastore"
+	dsq "github.com/ipfs/go-datastore/query"
+	"go.opentelemetry.io/otel/attribute"
+	otel "go.opentelemetry.io/otel/trace"
+)
+
+// New returns a new traced datastore. All datastore interactions are traced.
+func New(ds ds.Datastore, tracer otel.Tracer) *Datastore {
+	return &Datastore{ds: ds, tracer: tracer}
+}
+
+// Datastore is an adapter that traces inner datastore interactions.
+type Datastore struct {
+	ds     ds.Datastore
+	tracer otel.Tracer
+}
+
+var (
+	_ ds.Datastore           = (*Datastore)(nil)
+	_ ds.Batching            = (*Datastore)(nil)
+	_ ds.PersistentDatastore = (*Datastore)(nil)
+	_ ds.TxnDatastore        = (*Datastore)(nil)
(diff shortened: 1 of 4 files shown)


trace/trace.go
```diff
@@ -0,0 +1,228 @@
...
+		return dstore.Check(ctx)
+	}
+
+	return nil
+}
+
+// NewTransaction implements the ds.TxnDatastore interface.
+func (t *Datastore) NewTransaction(ctx context.Context, readOnly bool) (ds.Txn, error) {
+	ctx, span := t.tracer.Start(ctx, "NewTransaction", otel.WithAttributes(attribute.Bool("readOnly", readOnly)))
+	defer span.End()
```
Stebalien, line 222: Is there any way to associate gets/puts within the transaction with the transaction's span?

trace/trace.go
```diff
@@ -0,0 +1,228 @@
...
+	_ ds.CheckedDatastore    = (*Datastore)(nil)
+	_ ds.ScrubbedDatastore   = (*Datastore)(nil)
+	_ ds.GCDatastore         = (*Datastore)(nil)
+	_ io.Closer              = (*Datastore)(nil)
+)
+
+// Put implements the ds.Datastore interface.
+func (t *Datastore) Put(ctx context.Context, key ds.Key, value []byte) (err error) {
+	ctx, span := t.tracer.Start(ctx, "Put", otel.WithAttributes(attribute.String("key", key.String())))
+	defer span.End()
```
Stebalien, line 42: We're not tracing errors anywhere. We probably want to call `span.RecordError(err)` for all errors here (it'll ignore nil errors).

Stebalien (commented):
Two nits, but this otherwise looks great!

trace/trace.go
dennis-tra: ah good point, I guess it also makes sense to set the status then:

```
	// RecordError will record err as an exception span event for this span. An
	// additional call to SetStatus is required if the Status of the Span should
	// be set to Error, as this method does not change the Span status. If this
	// span is not being recorded or err is nil then this method does nothing.
```

trace/trace.go
dennis-tra: I wouldn't know how to associate that :/ 

I guess the association would happen at the level where the transaction is created. We could carry around the span context in the `Txn` struct but then what would you do with the context that gets passed into transaction put/gets?

Stebalien: Yeah, fair enough. I'd just leave it as-is.

trace/trace.go
Stebalien: Well, that API is bonkers. But ok.

Stebalien (approved):

PR #210, opened 2023-10-17
feat: TxnDatastore namespace and keytransform wrappers

i-norden:
This PR introduces namespace and keytransform wrappers for the go-datastore `TxnDatastore`. This is in order to support a `GetMany` boxo blockstore implementation that leverages the `TxnDatastore` for batched reads which, in turn, is ultimately for supporting parallel traversal methods in go-hamt-ipld and go-amt-ipld.

boxo/blockstore PR that uses this: https://github.com/ipfs/boxo/pull/492

### DIFF
--- a/keytransform/txndatastore.go
+++ b/keytransform/txndatastore.go
@@ -19,12 +19,12 @@ func WrapTxnDatastore(child ds.TxnDatastore, t KeyTransform) *TxnDatastore {
 		panic("child (ds.TxnDatastore) is nil")
 	}
 
-	return &TxnDatastore{ds: Wrap(child, t), child: child, KeyTransform: t}
+	return &TxnDatastore{Datastore: Wrap(child, t), child: child, KeyTransform: t}
 }
 
 // TxnDatastore keeps a KeyTransform function
 type TxnDatastore struct {
-	ds    *Datastore
+	*Datastore
 	child ds.TxnDatastore
 
 	KeyTransform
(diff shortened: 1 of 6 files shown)


keytransform/txndatastore.go
```diff
@@ -0,0 +1,415 @@
...
+
+// Children implements ds.Shim
+func (d *TxnDatastore) Children() []ds.Datastore {
+	return []ds.Datastore{d.child}
+}
+
+// Put stores the given value, transforming the key first.
+func (d *TxnDatastore) Put(ctx context.Context, key ds.Key, value []byte) (err error) {
+	return d.child.Put(ctx, d.ConvertKey(key), value)
+}
```
hsanjuan: At first sight, re-implementing every method is not necessary.

The embedded datastore will provide them, that is why it is embedded right?. You just need to add NewTransaction() to this type.

hsanjuan (changes requested):

keytransform/txndatastore.go
i-norden: `Query()`, `prepareQuery()`, and `Batch()` also need to be re-implemented to wrap things with the `KeyTransform`'s `InvertKey()` and `ConvertKey()`.  I believe everything else (except `NewTransaction()`) is already falling through to the embedded `ds.TxnDatastore`.

i-norden: Ah I see I think we can make the embedded child datastore a `keytransform.Datastore` and then all methods except `NewTransaction` can fall through to that. Will try this, thanks!

i-norden: This works! In this manner we can support every method except `NewTransaction` on the `TxnDatastore` by falling through to an embedded `*keytransform.Datastore`. We still need to implement `Query` and `prepareQuery` on the `txnWrapper` in order to wrap the embedded `ds.Txn` with the `KeyTransform`.

hsanjuan:
@i-norden I don't see much has changed since last time?

i-norden:
> @i-norden I don't see much has changed since last time?

Ah apologies, I embedded but with a named field instead of anonymously so it still had the methods reimplemented but only to fall directly through. Fixed now!

i-norden:
Also embedded `ds.Txn` into `txWrapper`, but since most the methods need to apply `ConvertKey()` the only ones this saved us from reimplementing are `Commit()` and `Discard()`

hsanjuan:
@i-norden I'm not going to have much time in the next few days, but have a look to https://github.com/ipfs/go-datastore/pull/211. I think it does what you want. You don't need to re-implement anything, you just need to wrap stuff in other stuff. If I haven't done something completely wrong, just take that, improving the naming of things and test it a bit.

i-norden:
Thanks @hsanjuan! I appreciate the feedback. I'll take a look and test out that wrapper.

lidel:
@i-norden  is this PR still necessary or were you able to solve it with mentioned wrapper?

PR #238, opened 2025-07-25
feat: context datastore

Wondertan:
Batch and Txs allow grouping multiple writes or reads together to optimize speed or provide atomicity. However, there is no way to benefit from this grouping when you work across or outside the datastore interface boundaries. Example:
```go
  ctx := context.Background()
  ds := ds.NewMapDatastore()
  bs := blockstore.NewBlockstore(ds)

  // in this case, blocks are deleted independently, and there is no way to group them
  for _, key := range keys {
    bs.DeleteBlock(ctx, key)
  }
```

The context datastore solves the problem above, enabling the datastore underneath Blockstore to be wrapped. The wrapper then intercepts individual operations and reroutes them into the respective Read or Write. Example:

```go
  ctx := context.Background()
  ds := contextds.WrapDatastore(ds.NewMapDatastore())
  bs := blockstore.NewBlockstore(ds)

  batch, _  := ds.Batch(ctx)
  ctx = contextds.WithWrite(ctx, batch)

  // even if blocks are deleted independently, they are still grouped in the same batch
  for _, key := range keys {
    bs.DeleteBlock(ctx, key)
  }
  
  Batch.Commit()
```

### DIFF
--- a/context/context.go
+++ b/context/context.go
@@ -28,9 +28,12 @@ func GetRead(ctx context.Context) (datastore.Read, bool) {
 	return batch, ok
 }
 
-type key int
+type (
+	writeKeyTp int
+	readKeyTp  int
+)
 
 var (
-	writeKey key
-	readKey  key
+	writeKey writeKeyTp
+	readKey  readKeyTp
 )
--- a/context/context.go
+++ b/context/context.go
@@ -1,0 +1,36 @@
+package contextds
+
+import (
+	"context"
+
+	"github.com/ipfs/go-datastore"
+)
+
+// WithWrite adds a write batch to the context.
+func WithWrite(ctx context.Context, batch datastore.Write) context.Context {
+	return context.WithValue(ctx, writeKey, batch)
+}
+
+// GetWrite retrieves the write batch from the context.
+func GetWrite(ctx context.Context) (datastore.Write, bool) {
+	batch, ok := ctx.Value(writeKey).(datastore.Write)
+	return batch, ok
+}
+
+// WithRead adds a read batch to the context.
+func WithRead(ctx context.Context, batch datastore.Read) context.Context {
+	return context.WithValue(ctx, readKey, batch)
+}
+
+// GetRead retrieves the read batch from the context.
+func GetRead(ctx context.Context) (datastore.Read, bool) {
+	batch, ok := ctx.Value(readKey).(datastore.Read)
+	return batch, ok
+}
+
+type key int
+
+var (
+	writeKey key
+	readKey  key
+)
--- a/context/context_ds.go
+++ b/context/context_ds.go
@@ -1,0 +1,95 @@
+package contextds
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/ipfs/go-datastore"
+	"github.com/ipfs/go-datastore/query"
+)
+
+var _ datastore.Datastore = (*Datastore)(nil)
+var _ datastore.Batching = (*Datastore)(nil)
+var _ datastore.TxnDatastore = (*Datastore)(nil)
+
+// WrapsDatastore wraps around the given datastore.Datastore making its operations context-aware
+// It intercepts datastore operations routing them to the current Write or Read if one exists on the context.
+func WrapDatastore(ds datastore.Datastore) datastore.Datastore {
+	return &Datastore{
+		inner: ds,
+	}
+}
+
+// Datastore is a wrapper around a datastore.Datastore that provides context-aware operations.
+// See [WrapDatastore].
+type Datastore struct {
+	inner datastore.Datastore
+}
+
+func (ds *Datastore) Put(ctx context.Context, key datastore.Key, value []byte) error {
+	if write, ok := GetWrite(ctx); ok {
+		return write.Put(ctx, key, value)
+	}
+	return ds.inner.Put(ctx, key, value)
+}
+
+func (ds *Datastore) Delete(ctx context.Context, key datastore.Key) error {
+	if write, ok := GetWrite(ctx); ok {
+		return write.Delete(ctx, key)
+	}
+	return ds.inner.Delete(ctx, key)
+}
+
+func (ds *Datastore) Get(ctx context.Context, key datastore.Key) ([]byte, error) {
+	if read, ok := GetRead(ctx); ok {
+		return read.Get(ctx, key)
+	}
+	return ds.inner.Get(ctx, key)
+}
+
+func (ds *Datastore) Has(ctx context.Context, key datastore.Key) (bool, error) {
+	if read, ok := GetRead(ctx); ok {
+		return read.Has(ctx, key)
+	}
+	return ds.inner.Has(ctx, key)
+}
+
+func (ds *Datastore) GetSize(ctx context.Context, key datastore.Key) (int, error) {
+	if read, ok := GetRead(ctx); ok {
+		return read.GetSize(ctx, key)
+	}
+	return ds.inner.GetSize(ctx, key)
+}
+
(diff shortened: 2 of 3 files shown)


Wondertan:
This change allowed us to save minutes(`341.9s`  vs `129.4s`) of deleting(pruning) chain data with the badger datastore.

gammazero:
This is a very nice way to generically handle using batching and transaction functionality of an underlying database. This could replace the `PutMany` method in `blockstore.NewBlockstore`. That would mean that `NewBlockstore` would not need to take a `Batching` datastore, and could just take a plain `Datastore`. This seems more flexible because it allows the same approach to supporting transactions.


@lidel Now we need to decide if we want to support the generaic approach only and get rid of `PutMany` and similar, keep both, or something else.

@Wondertan An alternative solution to your delete problem is to add a `DeleteMany` method to `Blockstore` and then batch delete using the `Batching` interface of the internal `Datastore`, similar to how `blockstore.PutMany` works.

Wondertan:
The DeleteMany wouldnt work outside of Blockstore. The blockstore is an example I gave to be familiar for shipyard folks, but its not exactly my case, so DeleteMany wouldnt be sufficient.

Wondertan:
I made a discovery that I am currently confused about. The performance improvement results from this change are wrong. The results are actually measured with [hacky poc](https://github.com/celestiaorg/go-ds-badger4/commit/8af067b9d326849b25ea3672ee5c71018a147b0f) for this change. The final, supposed to be the production version in this PR, actually adds time overhead of ~20s compared to the baseline.

Wondertan:
False alarm. I triple checked, and there was a bug on the context datastore usage side, which made this change effectively a noop. After the fix, the results are on par with the hacky version I referenced earlier.

gammazero:
This is a modification to low-level interfaces that will require many potential changes. We need to take more time to consider the larger implications of this change. This is currently outside of the scope of our current maintenance workload agreed with the IPFS foundation in 2025, so we need to defer this until a later time. This can be discussed for inclusion in future dev grant work.

Wondertan:
@gammazero, I am a bit confused. How is this change may require changes? Its an independent feature that doesn't have any connections with any existing code in kubo/boxo/rainbow/libp2p/etc. It doesn't hurt to include this and then later on IPFS representatives can decide if they want to integrate this feature. Meanwhile, merging and releasing this would allow us to remove the ugly replace in go.mod across our repositories. If you are worry about maintenance overhead, I can reassure you that you can ping me and I will there to help with whatever issue there is with this code. However, I highly doubt there will be unless, you actually put efforts into integrating this.

gammazero:
@Wondertan It is not that this change itself requires any changes, but about the potential consequences leading to changes. What I mean by this is that we would need to decide how to recommend that implementers use the new context datastore, particularly with other competing APIs like `PutMany` and such. Do we convert our existing code to use the context datastore so that we provide the preferred way of doing things? We just have not had time to consider the indirect implications of having the context datastore.

Here are some of the issues we have not had sufficient time to consider:

High level
- Yes context wrappers are a pretty convenient way to add extensibility for anything that was going to have async or tracing behavior anyway (both of which tend to carry through contexts). It also really helps in minimizing changes across codebases (e.g. particularly for something super low level like go-datastore).
  - Note: This is one of the reasons why it'd be nice to move towards context-based sessions for data transfer
- The downsides are generally that:
  - The set of extensibility options and what they can do is hard to track / document because they're everywhere
  - You have to worry a bit about clobbering (e.g. what happens if WithWrite gets called twice?)
- There's some side awkwardness in Go in general where if you wrap a struct to layer on top of it you have to figure out which pieces to re-expose which makes things harder (e.g. datastores can have batches, but blockstores don't have batches, etc.)

Regarding the PR it batches reads + writes but leaves commits out of scope. Is this what everyone wants / expects?

Possible alternatives here look like:
- Exposing batching everywhere up the stack (Get/Put/DeleteMany in Blockstore, Blockservice, DagService,, ....)
- Let (perhaps even encourage with some docs or examples?) people use context hacks everywhere but don't put them in go-datastore
- Propose changes to the context datastore that make it usable for a wider variety of situations
  - e.g. right now it only covers separating a few read operations from some write ones and then bundling those together. Also it hides any "feature" extensions on the base datastore (there are some wrapper tools and examples in go-datastore that might serve as example here)

Given your interest in this, I am going to put this back on our queue for reconsideration. However, I hope this gives a better explanation of why I was saying this was out of scope, at least for now.

Wondertan:
Hey @gammazero, appreciate the elaboration. Generally, I agree that there should be a single canonical way of approaching things, and in this case, having `PutMany` and the context datastore may add ambiguity. Still, in this particular case, I don't think blocking the merging of this PR until we figure out a single, consistent way of doing things is warranted, mainly because of how huge go-datastore ecosystem is. Even if we all agree on a single approach that covers all the use cases, migrating everyone is an endeavor and has to be done step-by-step, where the first step could be merging this.

What I can say for sure is that the current set of options within the repo does not enable the optimization this PR targets. and simply adding `DeleteMany` [would not work](https://github.com/ipfs/go-datastore/pull/238#issuecomment-3135632611) either. While there were ways to solve this selfishly on our side, I decided to opt for a solution that benefits a broader go-datastore ecosystem. It may not be the perfect one, nor would I claim we reached a rough consensus on this being a single canonical solution, but it solves my problem and a large number of adjacent issues in a fairly flexible way, well tested, and production-ready. 

My proposal is to extract the discussion on seeking the single canonical way for bathing things into a dedicated issue and unblock this PR. I am happy to add input on the use case we envision on our side. 

P.S. One of many things I've learned from IPFS codebases is how to structure dependency graph coherently, so that replaces are never needed. Please, don't force a replace on us :pray:, we are drowning in them already for orthogonal reasons. I had one confused user today who couldn't figure out where to get `context` pkg in his project.

gammazero:
@aschmahmann We need to make a decision whether this is within the scope of code we want to maintain and support. We are currently not consuming this code ourselves (in kubo, boxo, rainbow, or IPFS cluster), so we need to consider our policy for accepting things like this into our codebase.

Maybe we should consider turning this into a draft PR until there is something in our software that uses this.

Wondertan:
@aschmahmann, do you think we could bring clarity about next steps here soon? Should we assume that IPFS Shipyard no longer accepts external contributions? Please let us know.

aschmahmann:
@Wondertan apologies for not getting to this earlier.

> Should we assume that IPFS Shipyard no longer accepts external contributions?

No, we accept contributions across the various repositories we maintain

---

As for this PR in particular, my understanding is the objections to including it are:
1. You are the only one using it
2. The only benefit in including it in go-datastore vs elsewhere is that perhaps someone else might discover it. i.e. "Meanwhile, merging and releasing this would allow us to remove the ugly replace in go.mod across our repositories." and "Please, don't force a replace on us 🙏" is not necessary this code could just live in your codebase with no negative ramifications. If there had been some feature, optionality knob, etc. here that would enable you to implement something in your codebase that's a different story.
3. We can see from the set of "helpful utility datastores" in this repository that the majority of them that were envisioned as "lots of people will probably use this" have not in fact ended up this way (you can see this via github search). This includes mount, delayed, retrystore, etc. so adding more in doesn't seem super valuable unless there are actually multiple parties that want to use it (i.e. if Celestia and any other party were interested and wanted to use go-datastore as a place to add docs/bugfixes/etc. that'd probably be fine).

All this being said you've been around the ecosystem enough that if you feel the objections above are unwarranted and this would be useful to people we're fine to take it and see how it plays out. Lmk if you'd still like us to merge and we'll do it.

Wondertan:
Yes, we would appreciate a merge here :pray:! Thank you @aschmahmann.

If this datastore causes any form of issue or brings maintenance overhead, don't hesitate to ping me on GH and we will handle it.

PR #239, opened 2025-07-30
feat(keytransform): support transaction feature

### LINKED ISSUE #237: Add support for TxnFeature for all standard datastore implementations
Txn provides atomicity, unlike Batches as per the interface definition. However, every provided implementation only supports Batches.

Wondertan:
Closes #237

### DIFF
--- a/keytransform/keytransform.go
+++ b/keytransform/keytransform.go
@@ -2,7 +2,7 @@ package keytransform
 
 import (
 	"context"
-	"fmt"
+	"errors"
 
 	ds "github.com/ipfs/go-datastore"
 	dsq "github.com/ipfs/go-datastore/query"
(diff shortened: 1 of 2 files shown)


keytransform/keytransform.go
```diff
@@ -224,6 +226,23 @@ func (d *Datastore) Batch(ctx context.Context) (ds.Batch, error) {
 	}, nil
 }
 
+func (d *Datastore) NewTransaction(ctx context.Context, readOnly bool) (ds.Txn, error) {
+	tds, ok := d.child.(ds.TxnDatastore)
+	if !ok {
+		return nil, fmt.Errorf("txn not supported")
```
gammazero: Prefer `errors.New` when no formatting. Also want to avoid abbreviations in message.
```suggestion
		return nil, errors.New("transaction not supported")
```

gammazero (approved):

keytransform/keytransform.go
Wondertan: de4fd81

